PR #15, opened 2023-12-29
feat(create-vuepress): add create-helper (close #1)

### LINKED ISSUE #1: [Feature request] Create helper

`create-vuepress` does not support VuePress 2 yet.

### Suggested solution

- Update `create-vuepress` 

### Alternative

_No response_

### Additional context

_No response_

Mister-Hope:


create/package.json
meteorlxy, line 1: Better to put it inside another directory. E.g. `tools/create-vuepress` or `packages/create-vuepress`. I think you also have some other packages to put inside it

meteorlxy (changes requested):

tools/create-vuepress/template/docs/.vuepress/config.js
```diff
@@ -0,0 +1,15 @@
...
+
+  title: 'VuePress',
+  description: 'My first VuePress Site',
+
+  theme: defaultTheme({
+    logo: 'https://vuejs.press/images/hero.png',
+
+    navbar: ['/', '/get-started'],
+  }),
+})
```
meteorlxy, line 15: Need to set bundler in template config file

meteorlxy (changes requested):

nruffing:
* Should we generate a `.gitignore` or try to append to an existing one?

* Should we add a typescript option?

These might be better as a future enhancement.

Mister-Hope:
typescript is not actually needed bro, all ts files are handled by eslint, and editorslike.vscode can use built-in ts lib to parse the project.

nruffing:
For typescript I was referring to adding an option to generate the vuepress site with *.ts files instead of *.js.  This would be similar to the option in the the create vue template using `pnpm create vue`.

nruffing (changes requested):

Mister-Hope:
cc @nruffing Could you please check this again?

tools/create-vuepress/src/flow/createPackageJson.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+export const createPackageJson = async ({
+  targetDir,
+  packageManager,
+  locale,
+  preset,
+  bundler,
+}: CreatePackageJsonOptions): Promise<void> => {
+  const packageJsonPath = join(targetDir, 'package.json')
+  // TODO: Update it
```
meteorlxy, line 38: You can make use of `npm view vuepress@next version` command to get the latest version

Mister-Hope: I am negative towards it, a lot of area in China can counter timeout at this stage.

Mister-Hope:
> For typescript I was referring to adding an option to generate the vuepress site with *.ts files instead of *.js. This would be similar to the option in the the create vue template using `pnpm create vue`.

I don't think this is needed. Vue need this because typescript need to be manually installed and `@vue/eslint-config-typescript`, `vue-tsc` and other tools could be handed. Also it's vue's job to show users how to write typescript in vue sfc, and how vue sfc are imported (and correctly resolved) in ts files.

However we do not need to show users the above things, and the docs config file can be converted to ts just by changing the extension name.  Expert users which are familiar with typescript and vue projects should have no issue converting the existing files to ts.

We already ask users a lot of questions, like language, bundler, preset, workflow, etc. While a lot of people may even not familiar with javascript. So under this point of view, rare people would benifit from this config.

PR #22, opened 2024-01-21
feat(theme-default): auto sidebar per route and e2e test setup (close…

### LINKED ISSUE #17: [Feature request] Auto Sidebar Per Page

The auto sidebar feature of the default theme is quite useful, especially in cases where the pages are autogenerated documentation from something like TypeDoc.

Currently you can only use the auto feature on all pages or manually specify the sidebar for every page. It would be great if you could turn the feature on per page as well. 

I was able to hack it to make it work for the [data-grid-vue](https://datagridvue.com) site using vue router nav guards but its not maintainable and would probably be pretty easy to implement in the theme unless the changes to not use the vue router are going to make the auto sidebar feature more difficult to implement in general.
https://github.com/nruffing/data-grid-vue/blob/main/vuepress/.vuepress/client.ts#L63

### Suggested solution

Implement passing 'auto' to a `SidebarConfigArray` for a given page in the theme's config object.

We could also change `SidebarConfigObject` to allow a string directly.

```ts
export type SidebarAuto = 'auto'
export type SidebarConfigObject = Record<string, SidebarConfigArray | SidebarAuto>
```

```ts
sidebar: {
  '/guide/': [
    {
      text: 'Guide',
      children: [
        '/guide/README.md',
        '/guide/columns.md',
        '/guide/sorting.md',
        '/guide/filtering.md',
        '/guide/paging.md',
        '/guide/more-coming-soon.md',
      ],
    },
  ],
  '/generated/': 'auto'
  '/dotnet-generated/': 'auto',
},
```

### Alternative

_No response_

### Additional context

_No response_

nruffing:

- [X] Provide a description in this PR that addresses **what** the PR is solving. If this PR is going to solve an existing issue, please reference the issue (e.g. `close vuepress/core#123`).

- [ ] Bug fix
- [X] New feature
- [ ] Other

### Description

The auto sidebar feature of the default theme is quite useful, especially in cases where the pages are autogenerated documentation from something like TypeDoc.

Currently you can only use the auto feature on all pages or manually specify the sidebar for every page. It would be great if you could turn the feature on per page as well.

This PR also contains the initial setup of the e2e tests for the default theme and a couple e2e tests of this new feature.

### Screenshots

**Before**

**After**

closes vuepress/core#17
closes vuepress/core#19

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.0.0-rc.2](https://github.com/vuepress/ecosystem/compare/v2.0.0-rc.1...v2.0.0-rc.2) (2024-01-26)
+
+### Features
+
+* **create-vuepress:** update packages version ([68d5005](https://github.com/vuepress/ecosystem/commit/68d5005b3d5eabc8e9e473bf498038edd3043901))
+
 # 2.0.0-rc.1 (2024-01-26)
 
 ### Bug Fixes
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "npmClient": "pnpm",
-  "version": "2.0.0-rc.1",
+  "version": "2.0.0-rc.2",
   "command": {
     "version": {
       "allowBranch": "main",
--- a/themes/theme-default/CHANGELOG.md
+++ b/themes/theme-default/CHANGELOG.md
@@ -3,6 +3,10 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [2.0.0-rc.2](https://github.com/vuepress/ecosystem/compare/v2.0.0-rc.1...v2.0.0-rc.2) (2024-01-26)
+
+**Note:** Version bump only for package @vuepress/theme-default
+
 # 2.0.0-rc.1 (2024-01-26)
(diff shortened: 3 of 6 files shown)


meteorlxy (changes requested):
@nruffing I have pushed some changes directly.

I found that some errors are thrown when clicking the auto sidebar. Please resolve it and better to add corresponding e2e testings for clicking them. Thanks!

image

Mister-Hope:
> @nruffing I have pushed some changes directly.
> 
> I found that some errors are thrown when clicking the auto sidebar. Please resolve it and better to add corresponding e2e testings for clicking them. Thanks!
> 
> image

I don't think this is because the code change, see https://github.com/vuepress/core/issues/1466 I belived this could be solved by https://github.com/vuepress/ecosystem/pull/24 naturally once https://github.com/vuepress/core/pull/1447 is merged upstream

meteorlxy:
I see. Let's release core first

Mister-Hope:
I would suggest adding the new option as 'heading` instead of 'auto', as it would be great if we can alos support 'structure', see https://theme-hope.vuejs.press/guide/layout/sidebar.html#generate-from-file-structure The docs sidebar is using "structure" keyword.

nruffing:
> I would suggest adding the new option as 'heading` instead of 'auto', as it would be great if we can also support 'structure', see https://theme-hope.vuejs.press/guide/layout/sidebar.html#generate-from-file-structure The docs sidebar is using "structure" keyword.
> 
> @nruffing

I see. That makes sense. I am going to make that change and reopen this PR.

PR #23, opened 2024-01-21
feat(theme-default): image alt option and set aria hidden on title if same as image alt (close #20)

### LINKED ISSUE #20: [Feature request] Brand Img Alt Text

Dequeue/Axe recommends not to include alt text on an image that is also already displayed next to it. A screen reader will then read it twice.
https://dequeuniversity.com/rules/axe/4.8/image-redundant-alt?application=AxeChrome

Currently the brand image in the default theme will also use the site title it displays as the alt text of the brand image.

### Suggested solution

I think we should add a theme option to specify the brand image alt text and default it to an empty string but still render the attribute like Dequeue recommends.

We will also need to make sure this works appropriately with screen readers when in a mobile resolution where the brand image is displayed but the title is hidden.

### Alternative

_No response_

### Additional context

_No response_

### LINKED ISSUE #21: [Bug report] Nav Landmark Needs Unique Role/Title/Label
### Description

Dequeue recommends all HTML landmarks have a unique role, title or aria-label. If the page is using the auto link page nav there are two nav elements on the page with the same landmark role.
https://dequeuniversity.com/rules/axe/4.8/landmark-unique?application=AxeChrome

### Reproduction

NA

### Used Package Manager

npm

### System Info

```shell
NA
```

nruffing:

- [ ] Bug fix
- [X] New feature
- [ ] Other

### Description

Dequeue/Axe recommends not to include alt text on an image that is also already displayed next to it. A screen reader will then read it twice.
https://dequeuniversity.com/rules/axe/4.8/image-redundant-alt?application=AxeChrome

Currently the brand image in the default theme will also use the site title it displays as the alt text of the brand image.

This PR adds an option to set the brand image alt text to be different from the site title and default to the site title if not set. If the logo alt text and site title are the same then `aria-hidden` is set on the site title so its not read by screen readers. 

Also, Dequeue recommends all HTML landmarks have a unique role, title or aria-label. If the page is using the auto link page nav there are two nav elements on the page with the same landmark role.
https://dequeuniversity.com/rules/axe/4.8/landmark-unique?application=AxeChrome

This PR adds unique labels to differentiate the navs.

### Screenshots

**Before**

**After**

closes #20 
closes #21


themes/theme-default/src/client/components/NavbarItems.vue
```diff
@@ -176,7 +176,11 @@ useUpdateDeviceStatus(
 </script>
 
 <template>
-  <nav v-if="navbarLinks.length" class="navbar-items">
+  <nav
+    v-if="navbarLinks.length"
+    class="navbar-items"
+    aria-label="site navigation"
```
Mister-Hope, line 182: I think this should be able to localized to another language

themes/theme-default/src/client/components/PageNav.vue
```diff
@@ -89,7 +89,11 @@ const nextNavLink = computed(() => {
 </script>
 
 <template>
-  <nav v-if="prevNavLink || nextNavLink" class="page-nav">
+  <nav
+    v-if="prevNavLink || nextNavLink"
+    class="page-nav"
+    aria-label="page navigation"
```
Mister-Hope, line 95: So is this

themes/theme-default/src/client/components/NavbarBrand.vue
```diff
@@ -48,6 +56,7 @@ const NavbarBrandLogo: FunctionalComponent = () => {
       v-if="navbarBrandTitle"
       class="site-name"
       :class="{ 'can-hide': navbarBrandLogo }"
+      :aria-hidden="navBarLogoAltMatchesTitle"
```
Mister-Hope, line 59: I dislike the change. An empty alt for logo image would hide the alerts, and that's what my theme is doing (yes I made a lot of improvements downstream at theme hope but not syncing them back the default theme).


For this pr let's discuss 2 things:

1. In what situation, the site builder needs to provide a name for logo ranther than the site title he manually defined? (Logo alt is needed to be set manually instead of an empty alt (`alt=""`)

2. What's the benefit for letting the reader to read the logo and omit the actual title text, instead of reading the title text and omitting the logo?

Mister-Hope (changes requested):
Answer the 2 question first before solving the other one.

I would say no to this PR at the moment, and I believe changing the logo alt to empty string is enough.

Mister-Hope:
Also one more thing I would like to declare, we should definitely provide the best a11y support for the visually impaired people and body impaired people, i.e.:

- Screen readers should read every page easily
- All major contents should be interacted with keyboard

And SEO is important for documentation sites, so that we'd better made every SEO improvement if we can.

---

And in my mind, it doesn't mean we should try to fit every "best practice" given by different tools and website. For most of the part, edge built-in issue reporter and lighthouse for Chrome is enough, and we may omit the rest or even some of the rules in the above 2 tools.

For adding `aria-label` for `<nav>`, it sounds good, but it has little improvement about he 3 things I mentioned earlier. Body impaired people doesn't need the label and so is search engine. For visually impaired people, they obviously understand the page-nav from the text in my theme:

image

And if you want to improve it, a unvisible label with text "prev" and "next" which can be picked by the screen reader is actually more useful then a `aria-label="page navigation"` (by default the screen reader won't read it!)

themes/theme-default/src/client/components/NavbarBrand.vue
nruffing: The current issue is the text is duplicated and thus read twice by screen readers.  I don't think the issue is about hiding the alert as much as the alerts actually point to bad behavior to some users (especially from Deque who has done more testing and research then any single person could around accessibility).

I decided to add the option for specifying different alt text so the site designer can decide what is best for them. I agree that most of the time the image alt text and the site title are going to be the same but not necessarily always. For example, I could see a situation where a company has a sub-title or slogan that they want read by a screen reader because its part of their logo but wouldn't want it displayed as the site title.

The benefit for reading the logo alt text instead of the title when they are the same is just that that was the easier and cleaner implementation. There is no difference to how its read by screen readers.

nruffing:
> Also one more thing I would like to declare, we should definitely provide the best a11y support for the visually impaired people and body impaired people, i.e.:
> 
> * Screen readers should read every page easily
> * All major contents should be interacted with keyboard
> 
> And SEO is important for documentation sites, so that we'd better made every SEO improvement if we can.
> 
> And in my mind, it doesn't mean we should try to fit every "best practice" given by different tools and website. For most of the part, edge built-in issue reporter and lighthouse for Chrome is enough, and we may omit the rest or even some of the rules in the above 2 tools.
> 
> For adding `aria-label` for `<nav>`, it sounds good, but it has little improvement about the 3 things I mentioned earlier. Body impaired people doesn't need the label and so is search engine. For visually impaired people, they obviously understand the page-nav from the text in my theme:
> 
> image
> And if you want to improve it, a unvisible label with text "prev" and "next" which can be picked by the screen reader is actually more useful then a `aria-label="page navigation"` (by default the screen reader won't read it!) A best practice which only benefits people when they are reading the html source code is obviously useless. The "best practice score" given by tools is far away comparing with impaired people's experience.

I disagree that built-in accessibility scanners is all that should be used. Deque does a vast larger amount of testing and research then Google or Microsoft does. They are setting the standards that Google and Microsoft use and so will appear to be pushing the standard, because they are. When they add best practices its because they think and are trying to make it a standard.

In the case of landmarks not being unique it is a problem for screen readers, not just a best practice. You can see an example [here](https://www.aditus.io/patterns/multiple-navigation-landmarks/) where when a screen reader is performing landmark navigation there are multiple landmarks that just say "navigation". This is before getting into any contents of each navigation instance. Yes, the nest page and previous page links should have screen readable labels but that is a separate concern.

themes/theme-default/src/client/components/NavbarBrand.vue
```diff
@@ -24,12 +24,20 @@ const navbarBrandLogo = computed(() => {
...
+    navbarBrandTitle.value.toLocaleUpperCase().trim() ===
+    navbarBrandLogoAlt.value.toLocaleUpperCase().trim(),
+)
 const NavbarBrandLogo: FunctionalComponent = () => {
   if (!navbarBrandLogo.value) return null
   const img = h('img', {
     class: 'logo',
     src: withBase(navbarBrandLogo.value),
-    alt: '',
+    alt: navbarBrandLogoAlt.value,
```
Mister-Hope, line 40: I still think reading out the title could be more natural, remove the aria-hidden on title and provode an emty alt unless an alt for navlogo as manually provided plz.

Mister-Hope (changes requested):

Mister-Hope:
Also I believe yoh should send a pr to docs repo.

themes/theme-default/src/client/components/NavbarBrand.vue
nruffing: Done, working on docs now.

nruffing:
Actually, now I remember what the big advantage was having the logo alt read instead of the site title. When viewing the site in a mobile width, the site title is not displayed and thus hidden by screen readers. By having the alt text always on the logo we avoid having to address that scenario explicitly.

nruffing:
Documentation PR created at https://github.com/vuepress/docs/pull/10

Mister-Hope:
> @Mister-Hope Actually, now I remember what the big advantage was having the logo alt read instead of the site title. When viewing the site in a mobile width, the site title is not displayed and thus hidden by screen readers. By having the alt text always on the logo we avoid having to address that scenario explicitly.

Good point, that makes sense

Mister-Hope (approved):

PR #30, opened 2024-01-27
feat(theme-default): heading-sidebar

### LINKED ISSUE #17: [Feature request] Auto Sidebar Per Page

nruffing:

- [X] Provide a description in this PR that addresses **what** the PR is solving. If this PR is going to solve an existing issue, please reference the issue (e.g. `close vuepress/core#123`).

- [ ] Bug fix
- [X] New feature
- [ ] Other

### Description

The auto sidebar feature of the default theme is quite useful, especially in cases where the pages are autogenerated documentation from something like TypeDoc.

Currently you can only use the auto feature on all pages or manually specify the sidebar for every page. It would be great if you could turn the feature on per page as well.

This PR also contains the initial setup of the e2e tests for the default theme and a couple e2e tests of this new feature.

### Screenshots

**Before**

**After**

closes #17


.eslintrc.cjs
```diff
@@ -39,5 +39,9 @@ module.exports = {
         'vue/one-component-per-file': 'off',
       },
     },
+    {
+      files: ['**/e2e/**/*.cy.ts', '**/e2e/cypress/**/*.ts'],
+      extends: 'plugin:cypress/recommended',
+    },
```
Mister-Hope, line 45: See line 24-26. we already have that

e2e/docs/.vuepress/config.ts
```diff
@@ -69,6 +69,7 @@ export default defineUserConfig({
           ],
         },
       ],
+      '/heading-sidebar/': 'heading',
```
Mister-Hope, line 72: Could you change the strcture to:

```
sidebar/
  auto/
  auto.md
  config/
    1.md
    2.md
  README.md
```

I believe this could be better, thanks.

themes/theme-default/package.json
```diff
@@ -56,6 +56,9 @@
     "vue": "^3.4.15",
     "vue-router": "^4.2.5"
   },
+  "devDependencies": {
+    "vitest": "^1.1.0"
+  },
```
Mister-Hope, line 61: I am not sure, but I think declaring vitest at workspace root should be enough? Is there any necessary to declare here?

themes/theme-default/src/client/composables/useSidebarItems.ts
```diff
@@ -195,11 +195,14 @@ export const resolveMultiSidebarItems = (
...
 
+  if (matchedSidebarConfig === 'heading') {
+    return resolveAutoSidebarItems(page, sidebarDepth)
+  }
   return resolveArraySidebarItems(
     page,
     router,
     path,
-    matchedSidebarConfig,
+    matchedSidebarConfig as SidebarConfigArray,
```
Mister-Hope, line 205: Hi, `matchedSidebarConfig ` should be type-narrowed in line 198, what's the infer type here that you have to use a type cast?

Mister-Hope (changes requested):
Some tweaks is suggested.

.eslintrc.cjs
nruffing: Good catch, that was a merge side effect.

e2e/docs/.vuepress/config.ts
nruffing: To clarify, are you suggesting we remove the `heading-sidebar` folder and replace the `sidebar` folder with this structure?

Screenshot 2024-01-28 at 9 13 51 AM

themes/theme-default/package.json
nruffing: Removed.

themes/theme-default/src/client/composables/useSidebarItems.ts
nruffing: This is left over from when I had added a helper function to check if the config was `heading`/`auto` and the then the TS isn't able to determine that you already have determined that `matchedSidebarConfig` must be a `SidebarConfigArray`.  This ended up being the spot it was needed so I removed the helper function and now, you are right, the explicit type case is not longer needed.

nruffing:
I created a docs PR for these changes.  https://github.com/vuepress/docs/pull/9

e2e/docs/.vuepress/config.ts
Mister-Hope: Yes! but a auto folder and a auto.md, auto folder is for theme config, and auto.md is for frontmatter.

Mister-Hope (approved):

Issue #31, opened 2024-01-27
[Feature request] Add vuepress helper


Mister-Hope:
### Clear and concise description of the problem

Add some helper functions for developer.

They might be commonly used while developing plugins and themes, but not part of functionality of vuepress core.

### Suggested solution

A new `@vuepress/helper` package.

---

# Details

### Noop module

Proving `@vuepress/helper/noopModule` and `@vuepress/helper/noopComponent` for empty pkg and vue component.

This can be useful when developers want to do tree-shaking with `alias`.

### Bundler Related

We provide a lot of bundler related files, as it's likely for a plugin or theme to modify bundler config.

Writing codes directly can be annoying, see the following example:

```diff
-    // extends options of @vuepress/bundler-vite
-    if (app.options.bundler.name === '@vuepress/bundler-vite') {
-      bundlerOptions.vuePluginOptions ??= {}
-      bundlerOptions.vuePluginOptions.template ??= {}
-      bundlerOptions.vuePluginOptions.template.compilerOptions ??= {}
-      const isCustomElement = bundlerOptions.vuePluginOptions.template.compilerOptions.isCustomElement
-      bundlerOptions.vuePluginOptions.template.compilerOptions.isCustomElement = (tag) => {
-        if (isCustomElement?.(tag)) return true
-        if (tag === 'my-custom-element') return true
-      }
-    // extends options of @vuepress/bundler-webpack
-    if (app.options.bundler.name === '@vuepress/bundler-webpack') {
-      bundlerOptions.vue ??= {}
-      bundlerOptions.vue.compilerOptions ??= {}
-      const isCustomElement = bundlerOptions.vue.compilerOptions.isCustomElement
-      bundlerOptions.vue.compilerOptions.isCustomElement = (tag) => {
-        if (isCustomElement?.(tag)) return true
-        if (tag === 'my-custom-element') return true
-      }
+    addCustomElement(bundlerOptions, app, 'my-custom-element')
```

The main problem is that we need to ensure the parent scope of our config exists and we should not drop any existing config. So that's our helper goal.

Docs already exists at https://shared.vuejs.press/node/bundler.html (published as `vuepress-shared`)

### Excerpt Related

For seo feed or blogging, we need a concept called excerpt to describe a page main content using html, while VuePress2 remove this from core.

A good generator should do the following things:

1. removing Vue Components because they can not be rendered outside page.
[...]

PR #33, opened 2024-01-27
feat(helper): add helper package

Mister-Hope:
See #31


Mister-Hope:
@nruffing Could you please review it? I believe the code is at good quality, but need some suggestions about the API design, especially option names in params

tools/helper/src/node/page/excerpt.ts
```diff
@@ -0,0 +1,253 @@
...
+
+export interface PageExcerptOptions {
+  /**
+   * Excerpt separator
+   *
+   * 摘要分隔符
+   *
+   * @default "<!-- more -->"
+   */
+  excerptSeparator?: string
```
nruffing, line 20: I think just `separator` is fine since its part of the `PageExcerptOptions`.

I was trying to thing if there was a better word then separator but the best I could think of was `boundary` and I don't know that is really any better.

nruffing:
> @nruffing Could you please review it? I believe the code is at good quality, but need some suggestions about the API design, especially option names in params.
> 
> E.g.: I am using `preserveTitle`, `preserveFenceDom`, `isCustomElement` in excerpt options, but maybe they can have better names, for example using `separator` instead of `excerptSeparator`?

Is there a specific reason we dont want to allow Vue inside the excerpt or leverage Vue as the mechanism to implement the logic?  For example, why couldn't we have a component with a default slot that does this same behavior?

Mister-Hope:
For all of the 3 things, Vue component could be rendered.

1. For feed, absolutely we can not get Vue component rendered in feed readers.
2. For seo, this is similar. We only need the text in excerpt.
3. For blogging, unless we introduce vue compiler (which is 500kb - 4mb based on the features in use), we can not dynamically render a template string to dom. I believe in most situation, developer should use `v-html` directive and append the excerpt to dom via innerHTML, in this case we still need to remove Vue Components

nruffing:
> For all of the 3 things, Vue component could be rendered.
> 
> 1. For feed, absolutely we can not get Vue component rendered in feed readers.
> 2. For seo, this is similar. We only need the text in excerpt.
> 3. For blogging, unless we introduce vue compiler (which is 500kb - 4mb based on the features in use), we can not dynamically render a template string to dom. I believe in most situation, developer should use `v-html` directive and append the excerpt to dom via innerHTML, in this case we still need to remove Vue Components

Got it, that makes sense. Thank you for explaining the use cases.

Mister-Hope:
It would be great if you can provide some suggestions about the export name of these helpers and their options, I am not good at giving names 🥺 Feel free to give any ideas about changing any name or shape.

Mister-Hope:
> For all of the 3 things, Vue component could not be rendered.
> 
> 1. For feed, absolutely we can not get Vue component rendered in feed readers.
> 2. For seo, this is similar. We only need the text in excerpt.
> 3. For blogging, unless we introduce vue compiler (which is 500kb - 4mb based on the features in use), we can not dynamically render a template string to dom. I believe in most situation, developer should use `v-html` directive and append the excerpt to dom via innerHTML, in this case we still need to remove Vue Components

And another important one, we need to do hacky things to get Vue components that is imported locally in page.🫠 So if someone really needs this, he might need to build a new pipeline to achieve this, and maybe make the excerpt "a vue component" instead of a html fragment.

nruffing:
> It would be great if you can provide some suggestions about the export name of these helpers and their options, I am not good at giving names 🥺 Feel free to give any ideas about changing any name or shape.

I left a couple of comments related to naming and comments but overall it looks good.

I am actually more wondering if the packaging organization makes sense.  Do we really want a utility package of a bunch of random stuff? It just seems like it could grow to be a little out of control and maybe something like the excerpt functionality belongs in a package related to just SEO utilities?

Mister-Hope:
I am planning to migrate [`vuepress-plugin-blog2`](https://plugin-blog2.vuejs.press/) [`vuepress-plugin-feed2`](https://plugin-feed2.vuejs.press/) [`vuepress-plugin-seo2`](https://plugin-seo2.vuejs.press/) and [`vuepress-plugin-sitemap2`](https://plugin-sitemap2.vuejs.press/) into this repo in the near future under `@vuepress/xxx`

Anyway we need a util package to store the common method of these, excerpt is one of them and it is important (VuePress v1 core concerpt and also a core concerpt in VuePress 2 early stage)

I don't plan to add uncommon ones into the helper package.

Mister-Hope:
I am doing heavy refactor and improving the code quality in my repo in recent days, and once the helper package is ready, I will start shipping them.

Mister-Hope:
> I left a couple of comments related to naming and comments but overall it looks good.

BTW, I am still not seeing any comments here, are you still working on it, or did you forget to finish your review?

nruffing (approved):

nruffing:
> > I left a couple of comments related to naming and comments but overall it looks good.
> 
> BTW, I am still not seeing any comments here, are you still working on it, or did you forget to finish your review?

My bad, I never submitted the review.

tools/helper/src/node/page/excerpt.ts
Mister-Hope: separator is using at gray-matter, so I think it's good to inherit this option name.

PR #37, opened 2024-01-29
feat(plugin-sitemap): add sitemap plugin

Mister-Hope:
- [x] sitemap plugin
- [x] sitemap plugin docs
- [x] related e2e tests
- [x] add it in theme-default


Mister-Hope:
The plugin code is ready and if would be great if you can review it.

I believe most of the code should be nice and clean, excerpt for the `ensureHostName` function. Suggestions how to improve the code are welcomed.

PR #79, opened 2024-02-26
🐛 typo redirect-plugin css name

BTMuli:

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Description

Fix the css var name which is diffrent from theme & plugin-redirect,now they are the same.

### Screenshots

**Before**

**After**


BTMuli:
Since it is only component style differ from var&theme var,so it is more suitable to only change the component style?

I will change the code,wait a moment plz_(:з」∠)_

PR #107, opened 2024-04-12
fix(plugin-blog): useBlogCategory() can't return currentItems correctly

ShuJun-Junical:

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Description

In plugin-blog, when the current path is URLEncoded (e.g. the path contains Chinese), useBlogCategory() will not return currentItems, decodeURIComponent() is added in the judgment so that it can correctly handle Chinese paths.

### Screenshots

**Before**

**After**


Mister-Hope:
I would suggest using normalizePath from vuepress/shared on the right side

PR #111, opened 2024-05-03
feat: init watermark plugin

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

Integrate [watermark-js-plus](https://github.com/zhensherlock/watermark-js-plus) into VuePress。

This plugin can add watermark to the pages, you can choose between add watermark globally or on specific pages. You can also choose between add text watermark or image watermark.


Mister-Hope:
I think the docs should be detailed telling defineWatermarkConfig can used to add hooks  to watermark.

plugins/plugin-watermark/src/node/watermarkPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+export const watermarkPlugin = ({
+  pageFilter = () => true,
+  ...options
+}: WatermarkPluginOptions = {}): Plugin => {
+  options.global ??= true
+
+  return {
+    name: 'vuepress-plugin-watermark',
+    define: {
+      __VUEPRESS_PLUGIN_WATERMARK_OPTIONS__: options,
```
Mister-Hope, line 16: Use styles in other plugins

plugins/plugin-watermark/src/node/watermarkPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+const __dirname = getDirname(import.meta.url)
+
+export const watermarkPlugin = ({
+  pageFilter = () => true,
+  ...options
+}: WatermarkPluginOptions = {}): Plugin => {
+  options.global ??= true
+
+  return {
+    name: 'vuepress-plugin-watermark',
```
Mister-Hope, line 14: Wrong name

plugins/plugin-watermark/src/shared/options.ts
```diff
@@ -0,0 +1,42 @@
+import type { Page } from 'vuepress'
+import type { WatermarkOptions as _WatermarkRawOptions } from 'watermark-js-plus/dist/types/src/types/index.js'
+
+export type WatermarkOptions = Partial<_WatermarkRawOptions>
```
Mister-Hope, line 4: This should be watermark client options, and I am.actually suggesting only provide complicated ones.

In most.cases we should expect users to set options in node.js side.

plugins/plugin-watermark/src/shared/options.ts
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Enabled global watermark
+   * @default true
+   */
+  global?: boolean
+  /**
+   * Page filter
+   * @default () => true
+   */
+  pageFilter?: (page: Page) => boolean
```
Mister-Hope, line 33: Maybe filter like other plugins.

plugins/plugin-watermark/src/node/watermarkPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+    define: {
+      __VUEPRESS_PLUGIN_WATERMARK_OPTIONS__: options,
+    },
+    clientConfigFile: path.resolve(__dirname, '../client/config.js'),
```
Mister-Hope, line 18: move it after other lifecyles plz.

plugins/plugin-watermark/src/node/watermarkPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+import { getDirname, path } from 'vuepress/utils'
+import type { WatermarkPluginOptions } from '../shared/index.js'
+
+const __dirname = getDirname(import.meta.url)
+
+export const watermarkPlugin = ({
+  pageFilter = () => true,
+  ...options
+}: WatermarkPluginOptions = {}): Plugin => {
+  options.global ??= true
```
Mister-Hope, line 11: Add a logger for debug propose could be better as this could probably be a theme plugin.

plugins/plugin-watermark/src/client/config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineClientConfig } from 'vuepress/client'
+import type { ClientConfig } from 'vuepress/client'
+import { setupWatermark } from './composables/watermark.js'
```
Mister-Hope, line 3: We prefer to import from index.js

plugins/plugin-watermark/src/client/composables/watermark.ts
```diff
@@ -0,0 +1,111 @@
...
+
+const {
+  global = true,
+  delay = 500,
+  watermarkOptions: options,
+} = __VUEPRESS_PLUGIN_WATERMARK_OPTIONS__
+
+const userOptions = ref<WatermarkOptions>({})
+
+export function defineWatermarkConfig(
```
Mister-Hope, line 25: We prefer to add this in helper folder.

plugins/plugin-watermark/src/client/composables/watermark.ts
```diff
@@ -0,0 +1,111 @@
...
+      userOptions.value = value
+    })
+  } else {
+    userOptions.value = userConfig
+  }
+}
+
+export function setupWatermark(): void {
+  const frontmatter = usePageFrontmatter<{
+    watermark?: boolean | Omit<WatermarkPluginOptions, 'pageFilter' | 'global'>
```
Mister-Hope, line 47: A frontmatter type should be provided in share in case a theme or user needs it.

plugins/plugin-watermark/src/client/composables/watermark.ts
```diff
@@ -0,0 +1,111 @@
...
+const {
+  global = true,
+  delay = 500,
+  watermarkOptions: options,
+} = __VUEPRESS_PLUGIN_WATERMARK_OPTIONS__
+
+const userOptions = ref<WatermarkOptions>({})
+
+export function defineWatermarkConfig(
+  userConfig: MaybeRefOrGetter<WatermarkOptions>,
```
Mister-Hope, line 26: A maybe ref could be a good approach at a glance, maybe you can open prs to support this in other plugins.

plugins/plugin-watermark/src/client/composables/watermark.ts
```diff
@@ -0,0 +1,111 @@
...
+    let options = toValue(opts)
+    // Blind mode default alpha is 0.005
+    if (options.mode === 'blind') {
+      options = { globalAlpha: 0.005, ...options }
+    }
+
+    if (
+      options.image &&
+      !isLinkHttp(options.image) &&
+      options.image.startsWith('/')
```
Mister-Hope, line 94: this is confusing, I believe you may want to return if the image is relative

plugins/plugin-watermark/src/client/composables/watermark.ts
```diff
@@ -0,0 +1,111 @@
...
+      typeof frontmatter.value.watermark === 'boolean' ||
+      !frontmatter.value.watermark
+    ) {
+      return { ...options, ...userOptions.value }
+    }
+    return { ...options, ...userOptions.value, ...frontmatter.value.watermark }
+  })
+  const enabled = computed(() => {
+    if (global) return frontmatter.value.watermark !== false
+    return !!frontmatter.value.watermark
```
Mister-Hope, line 64: Boolean is prefered, let terser or other plugins do the job

plugins/plugin-watermark/src/shared/options.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Delay time when global watermark is not enabled
+   * @default 500
+   */
+  delay?: number
+
+  watermarkOptions?: WatermarkRawOptions
```
Mister-Hope, line 41: missing js doc

plugins/plugin-watermark/src/shared/options.ts
```diff
@@ -0,0 +1,42 @@
...
+   * Watermark parent selector
+   */
+  parent?: 'body' | (string & { __z_ignore?: never })
+}
+
+export interface WatermarkPluginOptions {
+  /**
+   * Enabled global watermark
+   * @default true
+   */
```
Mister-Hope, line 27: maybe zh descriptions as well

plugins/plugin-watermark/src/node/watermarkPlugin.ts
```diff
@@ -0,0 +1,29 @@
...
+
+  return {
+    name: 'vuepress-plugin-watermark',
+    define: {
```
Mister-Hope, line 15: we prefer spacing between these

Mister-Hope (commented):
need some tweaks in shared folder, make sure they should be used at both side.

For example the plugin option containing pageFilter is a node side one, then it should be node/options.ts

This ensures that these kind of types can not be imported in wrong side.

plugins/plugin-watermark/src/client/composables/watermark.ts
pengzhanbo: image may be encoded in base64 and may not always be provided with a link.

plugins/plugin-watermark/src/client/composables/watermark.ts
pengzhanbo: I will add it to my to-do list.

plugins/plugin-watermark/src/client/composables/watermark.ts
Mister-Hope: the problem is that 94 is just a stricter check than 93, if 94 is santisty, there is always 93

Mister-Hope:
LGTM, I leave a comment at a check of image, check that, maybr you want to mean or?

plugins/plugin-watermark/src/client/composables/watermark.ts
pengzhanbo: The function `isLinkHttp` checks for `https://` and `//`, but it does not always start with `/`.

Mister-Hope (approved):
Almost LGTM, I will make some tweaks later

Mister-Hope:
Currently this PR is blocked by the latest commit at main. I am still figuring out that.

Mister-Hope:
BTW, could you add some e2e tests?

pengzhanbo:
> BTW, could you add some e2e tests?

When I run `pnpm e2e:dev` locally, I noticed that the `webServer.command` in `playwright.config.ts` is not being executed at all. Playwright starts running the tests without waiting for `http://127.0.0.1:9080` to return `200 OK`. I have to run `pnpm docs:dev` first before `pnpm e2e:dev` runs correctly.

Mister-Hope:
> > BTW, could you add some e2e tests?
> 
> When I run `pnpm e2e:dev` locally, I noticed that the `webServer.command` in `playwright.config.ts` is not being executed at all. Playwright starts running the tests without waiting for `http://127.0.0.1:9080` to return `200 OK`. I have to run `pnpm docs:dev` first before `pnpm e2e:dev` runs correctly.

yes, anywhere is containing outdated devs and I replaced with http-server,  However I haven't made things to work yet,  The ci is still falling in main

Mister-Hope (changes requested):
Fix e2e tests plz

pengzhanbo:
> Fix e2e tests plz

Done

Mister-Hope:
@pengzhanbo Hi, I rebuild the plugin, and you are welcomed to review this for me.

PR #117, opened 2024-05-07
perf(plugin-watermark): optimize watermark redraw when switching pages

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

- Optimizing the display of watermarks when switching pages:
  - Transition from a page with a watermark to a page without a watermark.
  - Transition between two different watermark pages.

- Adding default watermark presets.


pengzhanbo:
> I try to fix the globalAlpha and other impack you introduce, you may have a deeper check before we merge it.

It's OK.

PR #123, opened 2024-05-11
feat!: remove external-link-icon

Mister-Hope:


themes/theme-default/src/node/defaultTheme.ts
```diff
@@ -33,6 +32,13 @@ export interface DefaultThemeOptions extends DefaultThemeLocaleOptions {
    */
   hostname?: string
 
+  /**
+   * Whether show external link icon
+   *
+   * @default true
+   */
+  externalLinkIcon?: boolean
```
meteorlxy, line 40: Maybe put it to `DefaultThemeLocaleData`?

Mister-Hope: I think the selector is a little bit complicated for people to override if we add another parent class to mark a state according to locales. And I think in all cases people should not need different behavior in different locales.

So I made the style file an alias, and only the root option works. That means it has to be write here.

PR #125, opened 2024-05-11
perf(plugin-back-to-top): optimize backToTop feature

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

Adjust the backToTop button, add scrolling progress percentage text, and optimize the interaction effect.

### Screenshots

**Before**

image

**After**


Mister-Hope:
I'm not pretty sure about your contribution. I already have the percent computed defing the scroll precent,  Why should and other similar variable is introduced?

Also there is one more thing to be discussed,  In what situation shall people care about the actual and exact percentage that a page has been scrolled?

A progress bar around the icon should have been enough for this situation.

PR #132, opened 2024-05-13
feat: add plugin-markdown-image

Mister-Hope:


Mister-Hope:
@pengzhanbo  I am using plugin options for selectors and write dynamic style files.

Any better approach?

pengzhanbo (dismissed):

pengzhanbo:
> @pengzhanbo I am using plugin options for selectors and write dynamic style files.
> 
> Any better approach?

Using built-in styles is fine, but it can also be offered as an optional method. Users can choose not to use built-in styles and instead customize them by providing style templates in document form.

Mister-Hope:
Is there any limitation with the current behavior?

pengzhanbo (approved):

PR #138, opened 2024-05-15
feat(plugin-docsearch): improve search result item

Mister-Hope:


pengzhanbo (dismissed):

fdevans:
I think that this change has broken how we use search with Algolia.  It seems now we are seeing duplicate base URLs.  Can we get some description or context of what this change is and what we might need to adjust as a result?

Mister-Hope:
This makes every record links to be a path name instead of a full url.

And these pr helps where same document is deployed in multiple places and even in multiple base folders.

For example, if you index https://example.com/ , while it's also deployed in  https://example.github.io/docs/,  You will see full urls of example.com in both site,  And this is not definitely the  Behavior in the secondsite as it is using router push to navigate to its own pages.

After this pr, a page indexed as /a.html will be /a.html and /docs/a.html  in both sites, and its click behavior will both be `router.push("/a.html")`

fdevans:
We are using variables as part of the start_urls to index two subfolders `/docs/` and `/5.4.0/` (which is a version of those docs) and corresponding facets to have indexes for each subfolder.  Our Algolia index entries already have the proper URL and this is now adding a 2nd base to the URL and breaking the navigation.

Screenshot 2024-06-20 at 10 06 38 AM

You can see the URL in the screenshot has the proper `/docs/` entry but the result from this change gives us 
`"https://docs.rundeck.com/docs/docs/manual/content-converters/properties.html"`

Since our version dynamic can have different content for each previously released version.  Is there a config or something I'm missing to get this to work properly?

fdevans:
Here is our start URLs config that achieves this.

`  "start_urls": [
    {
      "url": "https://docs.rundeck.com/(?P<version>.*?)/",
      "variables": {
        "version": [
          "docs",
          "5.4.0"
        ]
      }
    }
  ],`

Our Base config is also dynamic as we continue to host older versions of the site

`  base: `/${setup.base ? setup.base + '/' : ''}`,`

the `setup.base` variable is provided as an environment variable during build with a hard coded backup in a local file.

I'm trying to find a way to work through this.  Ideas would be:
- Use the *url_withtout_variable* with this new functionality as that looks to be the format it's expecting.
- Disable this new functionality an use the `url` as provided in the search index.

fdevans:
So I think the concern here is the code assumes that the indexes are made with a root folder site and that all subfolder sites are the same content.  That's not the case for our use.

I've had to revert to earlier versions of the libraries for now.  I can run tests on updates or changes if we can figure out something.

Mister-Hope:
>  the indexes are made with a root folder site 

Nope, theme-hope is crawling `https://vuepres-theme-hope.github.io/v2/`,and everything works fine

> and that all subfolder sites are the same content. 

Yes, as if contents are different, you probably need 2 of them.

fdevans:
I can file a bug somewhere else if that's better, but it does duplicate the base in our setup.

PR #145, opened 2024-05-16
feat: support markdown options in markdown plugins

Mister-Hope:


pengzhanbo:
Removed `plugin-markdown-image.d.ts` and `plugin-markdown-math.d.ts`, and migrated to `types.d.ts`, as well as exported it by name.

Users can import types in the following ways：

```ts
/// <reference types="@vuepress/plugin-markdown-image/types" />
/// <reference types="@vuepress/plugin-markdown-math/types" />

export default {
  markdown: {
    image: {},
    math: {},
  }
}
```

Mister-Hope:
I am expecting this type can be automatically loaded when users are importing those plugins.

Can we do that by importing the reference inside the node entrance?

pengzhanbo:
> I am expecting these types can be automatically loaded when users are importing those plugins.
> 
> Can we do that by importing the reference inside the node entrance?

Currently, this is the only way to ensure the correct extension of typescript types.

Mister-Hope:
Cc @meteorlxy , So I think we might need a string unknown for markdown, or a new generic for defineUserConfig

plugins/plugin-markdown-image/tsconfig.build.json
```diff
@@ -2,7 +2,10 @@
   "extends": "../../tsconfig.build.json",
   "compilerOptions": {
     "rootDir": "./src",
-    "outDir": "./lib"
+    "outDir": "./lib",
+    "paths": {
+      "@vuepress/markdown": ["./src/types.d.ts"]
```
meteorlxy, line 7: It would be verbose to require this config. What about putting the `declare module` clause inside `options.ts`?

PR #152, opened 2024-05-21
feat: optimize the code block copy

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

- Generate HTML text for the copy button in the `plugin-prismjs`/ `plugin-shiki`, and insert it into the code block.
- `plugin-copy-code` only provides functionality for the copy button and default styles.


Mister-Hope (changes requested):
default copy class is too easy, and would probably affected by other styles.

Mister-Hope (changes requested):
It seems that you remove the `showInMobile` option.

My suggestions is that we should add it back.

tools/highlight-helper/src/node/copy-code-button/createCopyCodeButtonRender.ts
```diff
@@ -0,0 +1,45 @@
...
+  const root = getRootLangPath(app)
+  const locales: LocaleConfig<CopyCodeButtonLocaleOptions> = {
+    // fallback locale
+    '/': userLocales['/'] || copyCodeButtonLocales[root],
+  }
+
+  getLocalePaths(app).forEach((path) => {
+    locales[path] =
+      userLocales[path] || copyCodeButtonLocales[path === '/' ? root : path]
+  })
```
Mister-Hope, line 35: I think you can reuse the `getLocaleConfig` loguc.

E.g.: when users are only setting `{ '/zh/': { title: '复制代码' } }` I believe your code will show a `Copied` message, while the original plugin can show a `已复制`.

Mister-Hope (changes requested):

Mister-Hope:
Also, I did not find any thing about `vp-copy-ignore` class in current codes, so what's this option for?

When users are using twosplash, can we still provide the correct content?

tools/highlight-helper/src/node/copy-code-button/createCopyCodeButtonRender.ts
pengzhanbo: I don't believe this is necessary. While this approach may ensure some content is displayed, it is not accurate. We should not be providing compatibility for developers' or users' configuration errors; instead, we should expose the issue to them in order to address the missing configuration problems.

Mister-Hope:
BTW, I still dislike this PR change, as it splits a whole copy-code functionality inside 2 plugins, and users do not benefit from it in most cases.

We don't need any SSG render or SEO with the copy button actually, we just need a client side a11y support.

The things I actually concern is that can the 2 highlighter plugin provide a client side transformer, guiding the copy code plugin to make some transform to get the correct code.

PR #157, opened 2024-05-27
fix: fix preWrapper plugin

Mister-Hope:
The preWrapper plugin should only wrap a `<div>` element when the fence result is a `<pre>` element

### DIFF
--- a/plugins/plugin-shiki/src/node/markdown/preWrapperPlugin.ts
+++ b/plugins/plugin-shiki/src/node/markdown/preWrapperPlugin.ts
@@ -23,7 +23,11 @@ export const preWrapperPlugin = (
 
     let result = rawFence(...args)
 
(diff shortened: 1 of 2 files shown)


plugins/plugin-shiki/src/node/markdown/preWrapperPlugin.ts
```diff
@@ -23,7 +23,7 @@ export const preWrapperPlugin = (
 
     let result = rawFence(...args)
 
-    if (!preWrapper) {
+    if (!preWrapper || !result.startsWith('<pre')) {
       // remove `<code>` attributes
```
Mister-Hope, line 27: Need some detailed code about what it's for.

Technically speaking, we need a note describe what  attributes are shiki adding and why should we remove it.

Also I think the logic is still incorrect here.

pengzhanbo (approved):

PR #158, opened 2024-05-28
feat(helper): add normalize export

Mister-Hope:
Provide a common normalize style for all themes.


tools/helper/package.json
```diff
@@ -31,6 +31,7 @@
     "./shared": "./lib/shared/index.js",
     "./noopComponent": "./lib/client/noopComponent.js",
     "./noopModule": "./lib/client/noopModule.js",
+    "./normalize": "./lib/client/styles/normalize.css",
```
pengzhanbo, line 34: I believe the `.css` suffix should not be omitted here; it should clearly indicate that this is a `css` file.

pengzhanbo (approved):

PR #159, opened 2024-05-28
chore: update repo structure

Mister-Hope:
This pr split plugins and their docs into subfolders:

image


Mister-Hope:
Mainly we should talk about grouping here, any suggestions? @pengzhanbo  @meteorlxy 

I want to make this be quickly merged to avoid conflicts that might be hard to resolve.

Mister-Hope:
Maybe the title could be better as `New Feature Plugins` and `Tool Plugins`, will make an update later.

pengzhanbo (approved):

PR #161, opened 2024-05-28
feat: add `:line-numbers=number` supports

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other


themes/theme-default/src/client/styles/code.scss
```diff
@@ -229,18 +229,12 @@ div[class*='language-'] {
       padding-top: 1.3rem;
       font-size: 0.875em;
       line-height: $line-height;
-      counter-reset: line-number;
 
       .line-number {
         position: relative;
         z-index: 3;
         user-select: none;
         height: #{$line-height}em;
```
Mister-Hope, line 237: can we just set it to the number?

Mister-Hope:
can we split these 2 changes?

themes/theme-default/src/client/styles/code.scss
pengzhanbo: Originally intended to use `counter-reset: line-number attr(data-start number)`,
but due to its poor compatibility, the current approach was chosen.

I could directly write the style into `.line-numbers` instead of using `data-start`.

pengzhanbo:
> can we split these 2 changes?

I moved the attribute migration to another PR.

themes/theme-default/src/client/styles/code.scss
```diff
@@ -229,7 +229,6 @@ div[class*='language-'] {
       padding-top: 1.3rem;
       font-size: 0.875em;
       line-height: $line-height;
-      counter-reset: line-number;
```
Mister-Hope, line 232: emm? I believe this should be preserverd

Mister-Hope (changes requested):

themes/theme-default/src/client/styles/code.scss
Mister-Hope: nope, I was talking about `height: #{$line-height}em;` to be `line-height: #{$line-height};`

themes/theme-default/src/client/styles/code.scss
Mister-Hope, line 232: 239-242 should be removed?

Mister-Hope (changes requested):

themes/theme-default/src/client/styles/code.scss
pengzhanbo: Do not remove, currently using `counter-increment` for line numbering.

Mister-Hope: I  think its ok for you to use actual numbers like vitepress.

Mister-Hope (changes requested):
Please resolve the conflicts

themes/theme-default/src/client/styles/code.scss
pengzhanbo: When writing numbers into the DOM, `plugin-seo` includes the numbers from `line-numbers` in the `og:description`, but it does not include the `code blocks content`, which seems a bit strange. Perhaps `plugin-seo` needs to make some adjustments.

Mister-Hope (approved):

PR #175, opened 2024-05-29
chore(plugin-shiki): tweak options

Mister-Hope:


plugins/markdown/plugin-shiki/src/node/shikiPlugin.ts
```diff
@@ -35,9 +24,9 @@ export const shikiPlugin = ({
     })
 
     md.use(highlightLinesPlugin)
-    md.use<PreWrapperOptions>(preWrapperPlugin, { preWrapper })
```
Mister-Hope, line 38: This is not necessary.

pengzhanbo (approved):

PR #182, opened 2024-05-30
 feat(plugin-shiki): add logLevel option 

Mister-Hope:


Mister-Hope:
I believe this is a better approach

PR #188, opened 2024-05-31
feat(theme-default)!: rename components

Mister-Hope:
- Add `VP` prefix to component names
- `vp-` prefix in class.
- move style inside component to make overriding easier

Should sync some coming changes about components from new default theme as well @pengzhanbo


tsconfig.json
```diff
@@ -10,7 +10,13 @@
       "@internal/themeData": [
         "./plugins/plugin-theme-data/src/client/themeData.d.ts"
       ],
-      "@theme/*": ["./themes/theme-default/src/client/components/*"]
+      "@theme/composables": [
+        "./themes/theme-default/src/client/composables/index.js"
+      ],
+      "@theme/*": [
+        "./src/client/*/index.js",
```
Mister-Hope, line 17:
```suggestion
```

themes/theme-default/src/node/defaultTheme.ts
```diff
@@ -62,6 +62,10 @@ export const defaultTheme = ({
             path.resolve(__dirname, '../client/components', file),
           ]),
       ),
+      '@theme/composables': path.resolve(
```
pengzhanbo, line 65: I don't think this is a good idea. I assume your intention is to support overriding source files just like `@theme/*.vue`. However, the files pointed by `@theme/composables` export local files within the theme. Once `@theme/composables` is overwritten, it means all exports will be overridden, which may not meet your expectations and could lead to more issues.

PR #189, opened 2024-05-31
feat(helper): add `getHeaders()`

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other


tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+  headers: MenuItem[],
+  levels?: HeaderLevels,
+): MenuItem[] {
+  if (levels === false) {
+    return []
+  }
+
+  const levelsRange = levels || 2
+
+  const [high, low]: [number, number] =
```
Mister-Hope, line 108: emmm high? low?

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+
+  headers = headers.filter((h) => h.level >= high && h.level <= low)
+  // clear previous caches
+  resolvedHeaders.length = 0
+  // update global header list for active link rendering
+  for (const { element, link } of headers) {
+    resolvedHeaders.push({ element, link })
+  }
+
+  const ret: MenuItem[] = []
```
Mister-Hope, line 123: wha ia ret?

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+  // eslint-disable-next-line no-labels
+  outer: for (let i = 0; i < headers.length; i++) {
+    const cur = headers[i]
+    if (i === 0) {
+      ret.push(cur)
+    } else {
+      for (let j = i - 1; j >= 0; j--) {
+        const prev = headers[j]
+        if (prev.level < cur.level) {
+          ;(prev.children || (prev.children = [])).push(cur)
```
Mister-Hope, line 133: prefer ??=

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export function resolveHeaders(
+  headers: MenuItem[],
+  levels?: HeaderLevels,
+): MenuItem[] {
+  if (levels === false) {
+    return []
+  }
+
+  const levelsRange = levels || 2
```
Mister-Hope, line 106: Can't this be a default value for levels?

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+  /**
+   * The children of the header
+   */
+  children: Header[]
+}
+
+export type HeaderLevels = false | number | [number, number] | 'deep'
+
+// cached list of anchor elements from resolveHeaders
+const resolvedHeaders: { element: HTMLHeadElement; link: string }[] = []
```
Mister-Hope, line 33: Any reasons to set a global state here?

Prefer to have this passed as arguments

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+  const [high, low]: [number, number] =
+    typeof levelsRange === 'number'
+      ? [levelsRange, levelsRange]
+      : levelsRange === 'deep'
+        ? [2, 6]
+        : levelsRange
+
+  headers = headers.filter((h) => h.level >= high && h.level <= low)
+  // clear previous caches
+  resolvedHeaders.length = 0
```
Mister-Hope, line 117: Any reasons to not set this as an empty array?

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+    const clone = h.cloneNode(true) as Element
+    clone.querySelectorAll(ignore.join(',')).forEach((el) => el.remove())
+    text = clone.textContent || ''
+  } else {
+    text = h.textContent || ''
+  }
+  return text.trim()
+}
+
+export function resolveHeaders(
```
Mister-Hope, line 98: Prefer arrow functions always

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,143 @@
...
+   * - `[number, number]: the first number is the minimum level and the second number is the maximum level.
+   * - `deep`: same as `[2, 6]`, which means all headings from `<h2>` to `<h6>` will be displayed.
+   *
+   * @default 2
+   */
+  levels?: HeaderLevels
+}
+
+export function getHeaders({
+  selector = '#vp-content :where(h1,h2,h3,h4,h5,h6)',
```
Mister-Hope, line 67: I still want this be ahigh support range selector,  Maybe creating with  Array.map

Mister-Hope (changes requested):

tools/helper/src/client/utils/getHeaders.ts
pengzhanbo: This is just a temporary variable used to store the result, whether named `res`, `list`, `ret`, or any other variable name, they are all interchangeable.

tools/helper/src/client/utils/getHeaders.ts
pengzhanbo: I removed it as it serves another purpose in `theme-next`, but extracting it to `helper` is pointless.

tools/helper/src/client/utils/getHeaders.ts
pengzhanbo: I don't think that overly flexible parameter types are necessary. As a selector, strings are already sufficient to meet the requirements, without assuming more possibilities.

Using `string[]` with `ignore` is because the nodes to be ignored may be isolated from each other without any inherent correlation.

tools/helper/src/client/utils/getHeaders.ts
```diff
@@ -0,0 +1,132 @@
...
+   *
+   * @default []
+   */
+  ignore?: string[]
+  /**
+   * The levels of the headers
+   *
+   * - `false`: No headers.
+   * - `number`: only headings of that level will be displayed.
+   * - `[number, number]: the first number is the minimum level and the second number is the maximum level.
```
Mister-Hope, line 55: The high and low variables  Seems to be the reverse logic here.

I understand that inthe document tree h 2 seems to be higher level of h3,  But for the numbers, two is lower than three. And that is usually what we compare of because we are setting 213, not the actual teg names,

So I still suggest  Adjusting those two variable names to rename them as min or max

tools/helper/src/client/utils/getHeaders.ts
Mister-Hope: I was trying to say that we can avoid using where by creating a long selector string from Array.map

Mister-Hope (changes requested):

tools/helper/src/client/utils/getHeaders.ts
pengzhanbo: As a default setting, I believe as long as it complies with the 'guideline', it should suffice. Through my practical testing, the compatibility of `:where` is actually better than what is stated in [caniuse](https://caniuse.com/mdn-css_selectors_where). Despite being marked as 'support unknown' in some mobile browsers, it is actually supported in browsers like `UC` and `quark`. As a utility function, the default setting for parameters should not be overdesigned. The final decision on how to pass them in should be left to the function caller, to be handled according to their specific scenario.

Mister-Hope (approved):

PR #198, opened 2024-06-05
feat: init guidelines

Mister-Hope:


docs/zh/themes/guidelines.md
```diff
@@ -0,0 +1,78 @@
+# 主题指南
+
+为了避免主题开发者和用户设置不必要的选项，我们制定了一套主题创建时应遵循的指南。
+
+## DOM 结构
+
+A theme must implement the following DOM structure:
```
pengzhanbo, line 7: It should be deleted in the zh docs.

themes/theme-default/src/client/styles/normalize.scss
```diff
@@ -1,72 +1,73 @@
...
 a {
-  color: var(--c-text-accent);
+  color: var(--vp-c-accent);
 }
 
 code {
-  background-color: var(--c-bg-light);
-  color: var(--c-text-lighter);
+  background-color: var(--vp-c-bg-alt);
+  color: var(--c-text-mute);
```
pengzhanbo, line 29: `--vp-c-text-mute`

plugins/features/plugin-copy-code/src/client/styles/vars.css
pengzhanbo, line 1: The icon has a problem and cannot be displayed correctly.

Mister-Hope: I noticed it too, so what's the problem? I already wasted an hour trying to find out the reason.

The only thing I did is deleting the size info, but it is still broken when I add it back.

Mister-Hope: Fuck, why stroke property does not accept a hex color value

pengzhanbo: It is supported, but the `#` in hex color needs to be handled as `%23`, so `#808080` should be rewritten as `%23808080`.

Mister-Hope: 🌚fine

PR #209, opened 2024-06-24
feat: add cache plugin

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

Vuepress has been criticized for its slow speed. 

This is because during the initialization phase, Vuepress needs to compile all markdown files. Although optimizations have been made in `vuepress/core` using `Promise.all()`, the cost of `createPage()` remains high, with the main expense coming from `markdown render()`. 

In practical use, as we add various markdown-it plugins, including the time-consuming code highlighting plugin, the compilation time remains reasonable when the number of markdown files is small. 

However, with a larger quantity, the time overhead increases significantly. It is unnecessary to recompile all markdown files every time Vuepress is started. For content that has not changed, skipping compilation can greatly improve startup speed.

When starting the development server to write markdown files, if there are multiple code blocks, typically we only update certain blocks within one modification cycle. However, when compiling the markdown file, it is still necessary to recompile all code blocks, which is unnecessary.

In fact, highlighting is also a relatively time-consuming operation, especially when using multiple transformers in `shiki` or even using `twoslash`, the time consumption is quite severe. Therefore, skipping the compilation of unchanged code blocks can greatly reduce the waiting time during hot updates.

This plugin does not solve the startup time consumption during the initial compilation, and it may even slightly increase the time taken for the first compilation. However, subsequent startups will benefit greatly from optimization, with larger projects seeing more noticeable effects.

The current approach is to back up all caches to a single cache file, while also writing them to the runtime memory, which might be more convenient for fast reading but could potentially consume too much memory.

Another option is to generate a separate cache file for each markdown file, using file system reading to achieve this. However, this introduces I/O overhead, and the individual I/O overhead for this part might be greater than the render overhead, possibly requiring additional computation to determine whether a file should use the cache.

### Screenshots

These startup time screenshots are from the `docs` of this repository.

**Unuse cache plugin**

image

**Use cache plugin, first Launch**

image

**Use cache plugin, Secondary Launch**

image


Mister-Hope:
Any reason why not to import a LRU package?

plugins/tools/plugin-cache/src/node/utils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export const hash = (data: string): string =>
+  createHash('md5').update(data).digest('hex')
+
+export const readFile = async <T = any>(filepath: string): Promise<T> => {
+  try {
+    const content = await fs.readFile(filepath, 'utf-8')
+    return JSON.parse(content) as T
+  } catch {
+    return {} as T
```
Mister-Hope, line 12: I would prefer null here for this api,.the needed of empty object might be better placed outside this function.

Mister-Hope:
I think we might need the memory usage comparison on some projects.

docs/.vuepress/configs/sidebar/en.ts
```diff
@@ -116,7 +116,12 @@ export const sidebarEn: SidebarConfig = {
     '/plugins/pwa/remove-pwa',
   ],
 
-  '/plugins/tools/': ['google-tag-manager', 'redirect', 'register-components'],
+  '/plugins/tools/': [
+    'google-tag-manager',
+    'redirect',
+    'register-components',
+    'cache',
```
meteorlxy, line 123: alphabetical ordering is preferred

docs/.vuepress/configs/sidebar/zh.ts
```diff
@@ -120,6 +120,7 @@ export const sidebarZh: SidebarConfig = {
     'google-tag-manager',
     'redirect',
     'register-components',
+    'cache',
```
meteorlxy, line 123: same here

plugins/tools/plugin-cache/src/node/utils.ts
```diff
@@ -0,0 +1,47 @@
+import { createHash } from 'node:crypto'
+import { fs, path } from 'vuepress/utils'
+
+export const hash = (data: string): string =>
```
meteorlxy, line 4: Is it possible to make use of the `hash` util from `vuepress/utils`?

plugins/tools/plugin-cache/src/node/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  try {
+    const tmp = path.join(cwd, 'tmp')
+    fs.writeFileSync(tmp, '{}', 'utf-8')
+    const start = performance.now()
+    readFileSync(tmp)
+    const end = performance.now()
+    fs.unlinkSync(tmp)
+    return end - start
+  } catch {
+    return 0.15
```
meteorlxy, line 45: What's the magic number for? For a default / fallback value, better to use a constant instead.

plugins/tools/plugin-cache/src/node/utils.ts
```diff
@@ -0,0 +1,47 @@
+import { createHash } from 'node:crypto'
+import { fs, path } from 'vuepress/utils'
+
+export const hash = (data: string): string =>
+  createHash('md5').update(data).digest('hex')
+
+export const normalizeFilename = (filename: string): string =>
+  hash(filename).slice(0, 10)
+
+export const readFile = async <T = any>(
```
meteorlxy, line 10: `fs.readJSON` and `fs.readJSONSync`?

plugins/tools/plugin-cache/src/node/utils.ts
pengzhanbo: I run IO tests with 100+ batches on my computer. The time-consuming falls in the range of approximately 0.1ms to 0.2ms, so I roughly chose 0.15 as a fallback option, which may not be used unless there are issues with read or write

plugins/tools/plugin-cache/src/node/utils.ts
pengzhanbo: I just noticed that `vuepress/utils` has `hash`, as long as the hash it generates is unique and stable, it is completely acceptable to use it.

Mister-Hope:
Is this PR ready for review yet?

pengzhanbo:
> Is this PR ready for review yet?

yes

docs/plugins/tools/cache.md
```diff
@@ -0,0 +1,43 @@
...
+
+By caching the `markdown render` during the initial startup of the VuePress development server, the plugin skips unnecessary `markdown render` on subsequent startups, thus speeding up the startup process.
+
+## Usage
+
+```bash
+npm i -D @vuepress/plugin-cache@next
+```
+
+It is recommended to place `cachePlugin` as the last item in the `plugins` configuration, as this can ensure maximum utilization of caching.
```
Mister-Hope, line 15: I think this should be better as a tip block

docs/plugins/tools/cache.md
```diff
@@ -0,0 +1,43 @@
...
+
+```ts
+import { cachePlugin } from '@vuepress/plugin-cache'
+
+export default {
+  plugins: [
+    // ... other plugins
+    cachePlugin({
```
Mister-Hope, line 23:
```suggestion
    // ... other plugins
    
    // place it in the end
    cachePlugin({
```

PR #231, opened 2024-07-28
feat: make sass optional and support sass-embedded

Mister-Hope:


pengzhanbo (approved):

Mister-Hope:
Webpack ci are all falling, could you help to investigate?

pengzhanbo:
> Webpack ci are all falling, could you help to investigate?

I investigated the issue for a while and can only confirm that the problem is caused by `plugin-pwa`, as the `vuepress build` completes but the process does not terminate.

I suspect that the issue lies with this line of code [pwaPlugin.ts#L71](https://github.com/vuepress/ecosystem/blob/main/plugins/pwa/plugin-pwa/src/node/pwaPlugin.ts#L71), as commenting it out allows the process to successfully terminate upon repeated execution. However, when building the project without the directories `e2e/.vuepress/{.cache,.temp,dist}`, the process does not terminate.

This confuses me; I am not clear about what happened to cause it.

Mister-Hope:
@meteorlxy any ideas?

PR #237, opened 2024-08-18
chore: use eslint v9

Mister-Hope:


Mister-Hope:
@meteorlxy Need help here.

1. There are many fail positives with `import/no-extraneous-dependencies`
2. We might need to allow `__PASCAL_CASE__` upstream
3. `declare module` is not working with eslint

@pengzhanbo Could you update the prismjs and shiki package to resolve the linter errors? I am not familar with those cdes.

Mister-Hope:
I have already fixed all the issues I could resolve.

meteorlxy:
1.There are many fail positives with import/no-extraneous-dependencies

Fixed. The plugins dir mapping was wrong. You only mapped to the category level, not the plugin level.

2. declare module is not working with eslint

Not only failed in eslint, but also failed in vscode. You might need more invetigation.

3. How should I bypass @typescript-eslint/no-use-before-define if 2 functions are referencing each other?

eslint-disable 😅

Mister-Hope:
> Not only failed in eslint, but also failed in vscode.

This is weird., everything works fine at my laptop:

image

Bothe typescript language service and eslint plugin are fine in my vscode.

What about yours? @pengzhanbo (note you might need to build first)

Mister-Hope:
Updated, this stopped working after I pull the latest commit and reload workspace, I will take further investigate

Mister-Hope:
@meteorlxy  I am not sure, but it seems that I was not able to add additional files here: 
image

pengzhanbo:
> What about yours

It can work normally.

pengzhanbo (approved):

PR #238, opened 2024-08-20
chore: bump deps

Mister-Hope:


Mister-Hope:
Also need some help here, I try to switch to modern api in Vite, but our usage about using `file:///` in `@forward` fails in Vite.

From the original PR https://github.com/vitejs/vite/pull/17754/files, it seems that a same importer with `legacy` is correctly applied to mordern api, so I am not sure how this issue happens

Mister-Hope:
I will try to split this one with:

1. deps bump and forcing to use the legacy one
2. a new pr using `sass-embedded` and `modern-api` by default.

The first one should be easy, and we will focus on the second one.

PR #239, opened 2024-08-21
feat: add plugin-markdown-hint

Mister-Hope:


plugins/markdown/plugin-markdown-hint/src/client/styles/hint.scss
```diff
@@ -0,0 +1,284 @@
+@use 'sass:color';
+@use 'icons';
+@use 'svg';
+
+.hint-container {
+  position: relative;
+  transition:
+    background var(--vp-ct),
+    border-color var(--vp-ct),
+    color var(--vp-ct);
```
pengzhanbo, line 10: I couldn’t find the definition of `var(--vp-ct)`.

plugins/markdown/plugin-markdown-hint/src/client/styles/hint.scss
```diff
@@ -0,0 +1,284 @@
...
+        }
+      }
+    }
+
+    p {
+      line-height: 1.5;
+    }
+
+    a {
+      color: var(--vp-tc);
```
pengzhanbo, line 71: I couldn’t find the definition of `var(--vp-tc)`.

plugins/markdown/plugin-markdown-hint/src/client/styles/hint.scss
Mister-Hope: This.should.be.unified to --vp-tc with guideline, but the main branch is not including it now.

PR #251, opened 2024-09-22
feat: add plugin-revealjs

Mister-Hope:


Mister-Hope:
@pengzhanbo 

This PR is almost ready with the following things that need to be specially reviewed:

1. To make the layout treeshakable, I define a alias and constanst in plugin, and I am wondering if we can make it more graceful.
2. The mouse scroll view is not working at the demo page, while the original demo from md-enhance (see https://plugin-md-enhance.vuejs.press/guide/content/revealjs/demo.html) is working fine. I already took time to investigate this, but failed to find out the problem.

docs/plugins/markdown/revealjs/README.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+```js {7} title=".vuepress/config.js"
+import { revealJsPlugin } from '@vuepress/plugin-revealjs'
+
+export default {
+  plugins: [
+    revealJsPlugin({
+      // Enable reveal.js
+      revealJs: true,
```
pengzhanbo, line 22: This configuration is not available in `RevealJsPluginOptions`.

docs/plugins/markdown/revealjs/README.md
```diff
@@ -0,0 +1,155 @@
...
+---
+
+<!-- slide2 -->
+
+---
+
+<!-- slide3 -->
+
+@slideend
+```
```
pengzhanbo, line 47: When I first browsed the document, I didn't understand the specific usage from the grammar; I only grasped it by looking at the source file of `demo.md` as an example. I suggest adding a concrete example here.

docs/plugins/markdown/revealjs/README.md
```diff
@@ -0,0 +1,155 @@
...
+- `blood`
+- `moon`
+
+For the appearance of each theme, see [Themes demo](themes.md).
+
+## Slide Layout
+
+By default, the plugin registers a global component `SlidePage` to render slides. In pages using this layout, you should only include a single slide syntax and no other contents to avoid rendering problems.
+
+If you do not need this layout or want to use another name, customize `layout` in plugin options.
```
pengzhanbo, line 72: As a developer, I can understand the description here, but for users, a sample containing `frontmatter.layout` is needed.
```md
---
layout: SlidePage
---
@slidestart
<!-- content -->
@slideend
```

docs/plugins/markdown/revealjs/themes.md
```diff
@@ -0,0 +1,267 @@
...
+
+```js [2-4|1-5]
+const add = (a, b) => {
+  if (typeof b === 'undefined') return a + 1
+
+  return a + b
+}
+```
+
+@slideend
```
pengzhanbo, line 25: This is a very good example, and perhaps the source code can be placed in a details container for users to view.
```md
::: detials source code
:::
```

pengzhanbo:
> * To make the layout treeshakable, I define a alias and constanst in plugin, and I am wondering if we can make it more graceful.

The current implementation seems fine, and further optimization would just be similar to the approach used in other plugins, like `(app) => prepareClientConfig(app)`.

Mister-Hope:
Any ideas with the second one?

PR #270, opened 2024-09-30
build: bump vuepress

Mister-Hope:


Mister-Hope:
Need help here, I am not sure why some lines in pwa plugin can not be type-narrowed correctly.

PR #283, opened 2024-11-02
feat(plugin-git): support changelog and improve performance

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

The current plugin executes three commands separately to obtain `createdTime`, `updatedTime`, and `contributors` for a single page. This results in even more command executions when `filepaths` contains multiple files, which is the main reason for the slow collection of git information.

Additionally, the `contributors` information often leads to duplicate statistics because the username in the local git configuration may not be the same as the username on the git hosting service, resulting in the same user having multiple usernames.

This PR makes the following updates and modifications:

- A single command is executed per file to collect all git information.

- Adds a `contributorConfig` preset for contributors' information, supporting **alias mapping**.

- Adds `avatar` and `url` information for contributors.

- Adds a `changelog` field to collect the change logs for the current page.

- Adds an `filter` field to control whether information is collected for pages globally.

- Adds `contributors` and `changelog` in `frontmatter` to control whether relevant information is collected for the current page.


docs/plugins/development/git.md
```diff
@@ -60,25 +60,129 @@ This plugin will significantly slow down the speed of data preparation, especial
...
+     * Contributor's username on the git hosting service
+     */
+    username: string
+    /**
+     * The alias of the contributor,
+     * Since contributors may have different usernames saved in their local git configuration
+     * compared to their usernames on the hosting service, In this case, aliases can be used to
+     * map to the actual usernames.
+     */
+    alias?: string[] | string
```
Mister-Hope, line 98: I think that providing a hook and letting users to make their own merge comparison is better

docs/plugins/development/git.md
```diff
@@ -60,25 +60,129 @@ This plugin will significantly slow down the speed of data preparation, especial
...
+
+- Default: `false`
 
 - Details:
 
-  A function to transform the contributors.
+  Whether to collect page changelogs or not.
 
-  The input is the contributors collected by this plugin, and the output should be the transformed contributors.
+### enabled
```
Mister-Hope, line 173: I prefer to call this `filter` and accept only function values.

The true and false seem like a redundant option to me.

plugins/development/plugin-git/src/node/types.ts
```diff
@@ -1,7 +1,189 @@
...
+  /**
+   * The url of the git repository, e.g: https://github.com/vuepress/ecosystem
+   *
+   * git 仓库的访问地址，例如：https://github.com/vuepress/ecosystem
+   */
+  repoUrl?: string
+
+  /**
+   * Commit url pattern
+   * Default: ':repo/commit/:hash'
```
Mister-Hope, line 128: it should be `@default`, explanation should be added with a `@description`

plugins/development/plugin-git/src/node/utils/getCommits.ts
```diff
@@ -0,0 +1,88 @@
...
+  } catch {
+    return []
+  }
+}
+
+export const parseRawCommits = (
+  rawCommits: string[],
+  filepath: string,
+): RawCommit[] => {
+  return rawCommits
```
Mister-Hope, line 43: Style issue

Mister-Hope (changes requested):

docs/plugins/development/git.md
pengzhanbo: The plugin still retains `contributors.transform` for handling `contributors`. This configuration mainly provides a simple way to preset contributor information.

plugins/development/plugin-git/src/node/types.ts
```diff
@@ -23,10 +214,71 @@ export interface GitData {
...
 
 export interface GitContributor {
   name: string
   email: string
   commits: number
+  avatar?: string
+  url?: string
+}
+
+export type GitType = 'bitbucket' | 'gitee' | 'github' | 'gitlab'
```
Mister-Hope, line 232: This could be better as KnownGitProvider?

plugins/development/plugin-git/src/node/types.ts
```diff
@@ -1,7 +1,198 @@
...
+   *
+   * - If the value is `false`, it will be ignored
+   * - If the value is `string[]`, it will be used as the list of extra contributors
+   */
+  contributors?: string[] | boolean
+
+  /**
+   * Whether to get the changelogs of a page
+   */
+  changelogs?: boolean
```
Mister-Hope, line 195: I am not sure, but I think I see the word changelog more often then changelogs

plugins/development/plugin-git/src/node/types.ts
```diff
@@ -1,7 +1,198 @@
...
+   * @default false
+   */
+  avatar?: boolean
+}
+
+export interface ContributorConfig {
+  /**
+   * Contributor's username on the git hosting service
+   *
+   * 贡献者在 git 托管服务中的用户名
```
Mister-Hope, line 76:
```suggestion
   * 贡献者在 Git 托管服务中的用户名
```

plugins/development/plugin-git/src/node/gitPlugin.ts
```diff
@@ -60,25 +41,63 @@ export const gitPlugin =
...
+            options,
+            gitType,
+            Array.isArray(frontmatter.contributors)
+              ? frontmatter.contributors
+              : [],
+          )
+        }
 
-          page.data.git.contributors = transformContributors?.(result) ?? result
+        if ((frontmatter.changelogs ?? changelogs) !== false) {
```
Mister-Hope, line 89:
```suggestion
        if (frontmatter.changelogs ?? changelogs ?? true) {
```

I'm reviewing the code on mobile phone, I think it's equivalent and you'd better double check it

plugins/development/plugin-git/src/node/gitPlugin.ts
```diff
@@ -60,25 +41,63 @@ export const gitPlugin =
...
 
+        if (filter && !filter(page)) return
+
+        const { frontmatter } = page
+
+        if (
+          !(frontmatter.changelogs ?? contributors) &&
+          !(frontmatter.contributors ?? changelogs) &&
+          !createdTime &&
+          !updatedTime
```
Mister-Hope, line 52: I am not understanding this check at first sight, and the second line seems weird. Some comments should be added here.

plugins/development/plugin-git/src/node/gitPlugin.ts
```diff
@@ -1,52 +1,33 @@
...
     contributors,
+    changelogs = false,
+    filter,
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
     transformContributors,
   }: GitPluginOptions = {}): Plugin =>
   (app) => {
     const cwd = app.dir.source()
     const isGitRepoValid = checkGitRepo(cwd)
+    const gitType = isGitRepoValid ? checkGitRepoType(cwd) : null
```
Mister-Hope, line 30: Should be gitProvider, and the function should be `referGitProvider`

Mister-Hope (changes requested):

Mister-Hope:
Title should be something like

`feat(plugin-git): support changelog and improve performance`

plugins/development/plugin-git/src/node/utils/checkGitRepo.ts
```diff
@@ -11,3 +12,26 @@ export const checkGitRepo = (cwd: string): boolean => {
     return false
   }
 }
+
+export const checkGitRepoType = (cwd: string): GitType | null => {
```
Mister-Hope, line 16: You are creating a git repo with `git init`, so repo type is a bad name to me.

Also I am wondering if we should check the first remote if origin doesn't exist?

pengzhanbo: If do it this way, we will need to execute multiple command lines here: first, use `git remote` to list all of them, then get the first one, and finally execute `git remote get-url xxx`.

Mister-Hope:
my bad, infer not refer 😂

Mister-Hope (approved):

PR #288, opened 2024-11-11
feat: add slimsearch plugin

Mister-Hope:


pengzhanbo:
It is recommended to adjust the minimum height of the modal box. The height change from **Searching...** to **No results found** after inputting content causes the modal box to shake.

Mister-Hope:
Could you please provide screenshots or just fix it for me?

also you can check the implementation. I think improvement could be made on my codes.

PR #297, opened 2024-12-07
fix(helper): `.sr-only` clip

typed-sigterm:

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Description

`clip` has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/clip), and `rect (0, 0, 0, 0)` should be `rect(0 0 0 0)`

### DIFF
--- a/tools/helper/src/client/styles/sr-only.scss
+++ b/tools/helper/src/client/styles/sr-only.scss
@@ -3,7 +3,7 @@
     position: absolute;
 
     overflow: hidden;
-    clip: rect (0, 0, 0, 0);
+    clip-path: rect(0 0 0 0);
 
     width: 1px;
     height: 1px;


Mister-Hope (approved):

labuladong:
Hi @Mister-Hope, I think `clip: rect(0 0 0 0);` should not be removed.  Some devices such as wechat browser don't support `clip-path`.

image

typed-sigterm:
@labuladong I tried that in WeChat, but it worked fine. It seems that my WeChat is using Chromium 130.

Also, I think the compatibility is okay - maybe we need a compatibility standard for the whole repo.

image

labuladong:
Oh, it failed in my WeChat and reproduced with https://developers.weixin.qq.com/miniprogram/dev/devtools/download.html

My workaround is adding this CSS manually.

Mister-Hope:
image

As far as the data shows: It is widely supported in any common browser. Only if where you are reporting, WeChat is using X5 (QQ browser core), that could be possible to fail. But that's out of concern here.

typed-sigterm:
There's another solution - just remove `clip-path` and `clip`, add `!important` to `.sr-only` rules.

Mister-Hope:
Update here, I tried WeChat devetools and reproduce what you are reporting:

image

It does say it does not support clip-path, but it's using chrome 86.

According to https://caniuse.com/mdn-css_properties_clip-path_basic_shape and real test on a chrome 55 with playwright, everything is working as expected.

Maybe you should raise an issue to wechat to team to as why this property is not supported as a real chrome 85 does. BTW try to reproduce youself in a chrome 85 first.

PR #308, opened 2024-12-16
feat(plugin-seo): improve `getCover`

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

Currently, the `plugin-seo` assumes that the cover image configuration field is a `string`, but the actual situation may be more complex. For example, if `cover` is of an object type, the current `getCover` may encounter errors that lead to build failures.

### DIFF
--- a/plugins/seo/plugin-seo/src/node/utils/getCover.ts
+++ b/plugins/seo/plugin-seo/src/node/utils/getCover.ts
@@ -1,7 +1,28 @@
-import { isLinkAbsolute, isLinkWithProtocol } from '@vuepress/helper'
+import {
+  isLinkAbsolute,
+  isLinkWithProtocol,
+  isPlainObject,
+  isString,
+} from '@vuepress/helper'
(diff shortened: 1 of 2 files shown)


Mister-Hope:
Can you explain why object format of cover and banner are used?

pengzhanbo:
For example, in the [theme plume](https://theme-plume.vuejs.press/guide/blog/#%E6%96%87%E7%AB%A0%E5%B0%81%E9%9D%A2%E5%9B%BE), you can configure the cover image in more detail.

pengzhanbo:
Implementing the `getter` in this plugin is another task that is not within the scope of this PR; this PR is solely an improvement to the default `getCover`.

Mister-Hope:
But the main problem is that I think assuming cover is string is fine. The plugin is designed to be out of box.

If you want to provide further information about the cover, setting it with `coverSettings` or `coverStyle` should be fine, so I would rather accept a PR supporting getter directly.

PR #319, opened 2024-12-22
feat: add plugin-icon

Mister-Hope:


Mister-Hope:
@pengzhanbo Please review this.

Note: We can support `iconify-icon` package as well, but I don't think adding full package support is needed.

1. The full package is too large, and it's fast enough to access icons from iconify apis.
2. Icon is definitely NOT important for SSR and SEO.

So I would suggest using the recommend way that iconify icon suggest, rather than supporting https://theme-plume.vuejs.press/guide/features/icon/#%E5%8A%A0%E8%BD%BD%E5%9B%BE%E6%A0%87

Mister-Hope:
Pinging @recoluan for your theme-reco as well

recoluan:
> Pinging @recoluan for your theme-reco as well

Good job!

Mister-Hope:
I decide to add this first. The next target is to support markdown grammar with it.

Then we can see if we can add built-in icons support for iconify.

For the grammar, I would prefer `::icon-name` instead of `:[icon-name]:`, any reason that you choose `:[xxx]:` @pengzhanbo ?

pengzhanbo:
@Mister-Hope 
`::icon-name` lacks a clear ending boundary, and icons often need to support settings for color, size, etc. Without a clear boundary, it is difficult to determine when to end.

`:[xx]:` references the emoji format `:xxx:`, allowing for more content to be written within `[]`. This can be expanded to the syntax `:[icon-name size/color]:` to support settings for size and color.

PR #348, opened 2025-01-24
feat(plugin-shiki): add built-in twoslash support

Mister-Hope:


Mister-Hope:
@pengzhanbo It seems that you already implement a good approach in your theme, please consider migrating the feature upstream.

pengzhanbo:
I see that the current PR uses `renderRich` as the rendering solution for `twoslash`.

In the early stages of my theme, I also used [`renderRich`](https://shiki.style/packages/twoslash#rendererrich), but the experience was not user-friendly. It was limited by the code block's `overflow: auto` setting; if the popup inside the code block was too large and exceeded the width or height of the code block, the popup would be partially obscured, and the code block would display a scrollbar.

I suggest using [`RenderFloatingVue`](https://shiki.style/packages/twoslash#rendererfloatingvue) as the rendering mode, which would provide a much better experience. Of course, this would require introducing more third-party dependencies like `floating-vue`, `mdast-util-from-markdown`, `mdast-util-gfm`, and `mdast-util-to-hast`.

If `twoslash` is to be integrated into `plugin-shiki`, having users manually install these dependencies related to `twoslash` would raise the barrier to usage. Should we consider making them default installed dependencies? Do you have any solutions regarding this?

Mister-Hope:
I would prefer optional deps plan, and if the deps are too many, we can consider providing an extra package which has them as deps, so that users can install a single package (e.g.: `@vuepress/shiki-twoslash`) to achieve

plugins/markdown/plugin-shiki/src/node/markdown/highlighter/createShikiHighlighter.ts
```diff
@@ -60,7 +67,19 @@ export const createShikiHighlighter = async ({
...
+  if (options.twoslash) {
+    const { transformerTwoslash: transformer } = await import(
+      '@vuepress/shiki-twoslash'
+    )
+
+    transformerTwoslash = await transformer({
+      twoslashOptions: isPlainObject(options.twoslash) ? options.twoslash : {},
+    })
+
+    await shikiSetup?.(highlighter)
```
Mister-Hope, line 81: 这里的位置对么？

plugins/markdown/plugin-shiki/src/node/prepareClientConfigFile.ts
```diff
@@ -80,15 +82,27 @@ export const prepareClientConfigFile = (
...
   let code = imports.join('\n')
 
-  if (setups.length) {
-    code += `\n
-export default {
-  setup() {
-    ${setups.join('\n    ')}
-  }
-}\n`
+  if (setups.length || enhances.length) {
```
Mister-Hope, line 97: 这里我可能还是更倾向用反引号加换行，可读性会好一点。

tools/shiki-twoslash/src/node/transformerTwoslash.ts
```diff
@@ -0,0 +1,97 @@
...
+import { createTwoslasher } from 'twoslash-vue'
+import { logger } from 'vuepress/utils'
+import { rendererFloatingVue } from './rendererFloatingVue.js'
+import { resolveTsPaths } from './resolveTsPaths.js'
+import type { VuePressTwoslashOptions } from './types.js'
+
+/**
+ * Create a Shiki transformer for VitePress to enable twoslash integration
+ */
+export const transformerTwoslash = async (
```
Mister-Hope, line 18: Could it be named something like `createTwoslashTransformer`?

tools/shiki-twoslash/src/client/twoslash.ts
```diff
@@ -0,0 +1,62 @@
+import Gt, { recomputeAllPoppers } from 'floating-vue'
+import type { App } from 'vue'
+import 'floating-vue/dist/style.css'
+
+const isMobile =
```
Mister-Hope, line 5: There might be a similar util in helper package.

tools/shiki-twoslash/src/client/twoslash.ts
```diff
@@ -0,0 +1,62 @@
...
+            (el as HTMLElement)?.classList?.contains('vp-code-tab'),
+          )
+        )
+          recomputeAllPoppers()
+      },
+      { passive: true },
+    )
+  }
+
+  app.use(Gt, {
```
Mister-Hope, line 30: What does Gt means here?

tools/shiki-twoslash/src/client/styles/twoslash.scss
```diff
@@ -0,0 +1,378 @@
+/* stylelint-disable selector-class-pattern */
```
Mister-Hope, line 1: Files like this may be updated with sass nesting so that it can have better readability

tools/shiki-twoslash/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "devDependencies": {
+    "@types/hast": "^3.0.4",
+    "vue": "^3.5.13"
+  },
+  "peerDependencies": {
+    "vuepress": "2.0.0-rc.19",
+    "shiki": "^2.3.2"
+  },
+  "peerDependenciesMeta": {
+    "shiki": {
```
Mister-Hope, line 61: Curious, why shiki is an optional peer here?

plugins/markdown/plugin-shiki/src/node/transformers/getTransformers.ts
```diff
@@ -86,3 +89,15 @@ export const whitespaceTransformer = (
 
   return [transformerRenderWhitespace({ position })]
 }
+
+const TWOSLASH_REGEXP = /\btwoslash\b/
+export const twoslashTransformer = (
```
Mister-Hope, line 94: Could it be totally moved to the new package?

plugins/markdown/plugin-shiki/src/node/markdown/highlighter/createShikiHighlighter.ts
pengzhanbo: 这里没有执行上的冲突。

tools/shiki-twoslash/src/client/twoslash.ts
pengzhanbo: The default export package name for the ESLint rule warning module. I think this rule can be turned off.

tools/shiki-twoslash/package.json
pengzhanbo: I added it when I first created the package, but after finishing the package, I realized it wasn't needed and can be deleted.

tools/shiki-twoslash/src/client/twoslash.ts
Mister-Hope: A eslint-disable-next-line could be used

plugins/markdown/plugin-shiki/src/node/transformers/getTransformers.ts
pengzhanbo: I don't think it's necessary; `@vuepress/shiki-twoslash` is an optional asynchronously loaded module, and it's better to maintain this method within `plugin-shiki`.

PR #356, opened 2025-02-18
chore(theme-default): add useData and prefer useData

Mister-Hope:


Mister-Hope:
Since the new default theme is using `useData`, I made a similar refactor on the old default theme.

It seems that the only benefit of the change is to reduce multiple imports and calls for developers in source code.

While the disadvantage is increasing size. Bundlers can not minify object keys, so all variable names will be left in the final output code.

After this PR, the size of the final production grows a little bit, and it can be even more if plugins and tool packages are following this change.

Honestly, since multiple imports in plugins are not common, I think we should not prefer `useClientData` in plugins, but for theme, the advantages and disadvantages seem kind of equal to me.

@pengzhanbo

pengzhanbo:
After compression, the destructuring syntax is optimized during the process. It will be transformed into: 

before:
```js
const { theme, frontmatter } = useData()
```
after:
```js
const {theme:a,frontmatter:b}=e()
```

The impact on the build size is minimal, but it will significantly enhance the development experience.

PR #369, opened 2025-02-24
feat: prefer useHeaders

Mister-Hope:


Mister-Hope:
The theme sidebar might need an animation. Can you help me add it? @pengzhanbo

PR #386, opened 2025-03-04
feat: add `codeBlockTitle` option support for highlighter, close #379

### LINKED ISSUE #379: [Feature request] Improve code block title appearance

The current title appearance is not good.

### Suggested solution

Something like the following:

https://vitepress.dev/guide/getting-started#the-config-file

### Alternative

_No response_

### Additional context

_No response_

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

close #379

### Screenshots

QQ_1741079980083


Mister-Hope:
Actually I would prefer the vitepress style, code title should be expanded to full width, otherwise it would be very confusing when combining with tabs or code tabs.

Mister-Hope (changes requested):
See comment

pengzhanbo:
> Besides, icon support could be added here, similar styles can be found in VitePress default theme, supporting by --icon

Regarding support for icons, I have added the attribute `data-title="xxx"` to the tags, but I do not intend to provide default styles for it; this part will be customized by theme developers or users through CSS extensions. It also supports passing a `render(title, code)` function to customize the rendering of code block titles.

Mister-Hope (approved):

PR #397, opened 2025-03-07
fix(plugin-markdown-hint): cleanup the `env` passed into `renderInline`

pengzhanbo:

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Screenshots

**Before**

**After**

### DIFF
--- a/plugins/markdown/plugin-markdown-hint/src/node/cleanMarkdownEnv.ts
+++ b/plugins/markdown/plugin-markdown-hint/src/node/cleanMarkdownEnv.ts
@@ -1,0 +1,12 @@
+import type { MarkdownEnv } from 'vuepress/markdown'
+
+export type ClearMarkdownEnv = MarkdownEnv & {
+  references?: unknown
+}
+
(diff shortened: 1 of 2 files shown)


plugins/markdown/plugin-markdown-hint/src/node/cleanMarkdownEnv.ts
```diff
@@ -0,0 +1,12 @@
+import type { MarkdownEnv } from 'vuepress/markdown'
+
+export type ClearMarkdownEnv = MarkdownEnv & {
+  references?: unknown
```
Mister-Hope, line 4: 这个是哪个插件提供的？

pengzhanbo: 这个是 [markdown-it](https://github.com/markdown-it/markdown-it/blob/master/lib/rules_inline/link.mjs#L83) 内部提供的。

PR #398, opened 2025-03-07
feat(plugin-git): add composables

Mister-Hope:


Mister-Hope:
TODO:

1. Update changelog component class name with vp prefix
2. Reduce page data size

RFC here:

1. git username is definitely not equalivent too github username, while the plugin is doing this as default behvaior. I think gavatar with email should be always perferred.
2. There is likely users contributing with both noreply and own emails, we need to special mention it in docs and provide guidelines for users merging contributors easily (probably provide utils functions to avoid writing complicated functions like this: https://github.com/vuepress-theme-hope/vuepress-theme-hope/blob/ce8916d7eee286c4238e811adf59c1d847a402cd/docs-shared/src/theme-wrapper.ts#L94-L127)
3. When detecting a noreply email, I think this should be removed in the commit info.

PR #430, opened 2025-04-12
refactor(helper): prefer @use

Mister-Hope:


pengzhanbo:
This is not the correct approach. I use `@import './vars.css'` because it allows for reuse. If  use the SCSS statement `@use 'vars'`, it will cause each compiled CSS file to inline the code from `vars` separately, leading to duplicate declarations, which is a mistake.

Mister-Hope:
I see, the import needs to be preserved.

PR #436, opened 2025-04-16
feat(plugin-git): sanitize commit messages

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

In previous versions, the submission information was rendered using `innerHTML`, but this could potentially contain illegal HTML structures. Therefore, it was modified to use `innerText` for rendering. However, this led to issues where content such as `#issue` within the submission information could not be rendered correctly.

Therefore, use `rehype` to sanitize commit messages, ensuring that the commit message HTML strings are clean and safe.

Ref:

[rehype](https://github.com/rehypejs/rehype)
[rehype-sanitize](https://github.com/rehypejs/rehype-sanitize)

### Screenshots

**Before**

**After**


Mister-Hope:
I am not familiar with these tool chain, but I have used dom-purify.

If HTML MUST be supported here, the sanitized process requires further discussion.

A general discussion can be found at https://github.com/walinejs/waline/blob/main/packages/server/src/service/markdown/xss.js

For example:

- form text like `<form>` `<input>` and media tags shall be forbidden, for commit msg, I believe only a small whitelist of tags shall be supported.
- Anything related to style must be forbidden, inducing ID, class and style
- Other attributes that uses may collect data or affecting site behavior shall also be blocked.

Mister-Hope (changes requested):
I believe some unit test shall be added in this PR

pengzhanbo:
Currently, in commit messages, a whitelist mechanism only supports the `a`, `code`, `em`, and `strong` tags. Among these, the `a` tag only supports the `href`, `target`, and `rel` attributes, and there is also validation for the value of the `href` attribute.

PR #449, opened 2025-04-30
feat(plugin-markdown-image)!: rebuild img size (close #445)

### LINKED ISSUE #445: [Feature request] markdown-image use better syntax

The default syntax of markdown img is to add additional data in the `img url` part in `img title`

This design, however, will result in a syntax that is not backward compatible with markdown syntax.

One example is that if I write `test2` and view it on other md platforms, there is a high probability that the image will not be displayed.

(I have tested vuepress and typora, both of which do not have the imgSize feature enabled, and they only display ordinary text or incorrect images instead of the specified images)

### Suggested solution

**I can do PR if you agree to my plan**.

The addition of additional information should be in the img title rather than the img url. This can achieve maximum downward compatibility.

A good example is the syntax in obsidian (built-in, not a plugin)

Image

I think this grammar can be added and suggest that it be recommended as the grammar in the document. (Old grammar can be retained.)
Meanwhile, this will also support users who create online documents from obsidian.

### Alternative

_No response_

### Additional context

_No response_

LincZero:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Description

close #445

### Screenshots

**Before**

**After**


docs/.vuepress/configs/plugins.ts
```diff
@@ -51,6 +51,7 @@ export const plugins = [
     figure: true,
     mark: true,
     size: true,
+    legacyImgSize: true,
```
Mister-Hope, line 54: This should be `legacySize`

docs/plugins/markdown/markdown-image.md
```diff
@@ -146,10 +190,10 @@ If the image is standalone in a line, wrapped or not wrapped by link, it will be
 - Details:
   Whether enable image size support.
 
-### obsidianSize
+### legacyImgSize
```
Mister-Hope, line 193: same

docs/plugins/markdown/markdown-image.md
```diff
@@ -28,6 +28,8 @@ export default {
       mark: true,
       // Enable image size
       size: true,
+      // Enable legacy image size (deprecated)
+      legacyImgSize: true,
```
Mister-Hope, line 32: I don't think it's worthy mention here.

docs/zh/plugins/markdown/markdown-image.md
```diff
@@ -28,6 +28,8 @@ export default {
       mark: true,
       // 启用图片大小
       size: true,
+      // 启用旧版图片大小 (不建议)
+      legacyImgSize: true,
```
Mister-Hope, line 32: Same

docs/zh/plugins/markdown/markdown-image.md
```diff
@@ -145,10 +189,10 @@ interface ImageMarkOptions {
 - 类型：`boolean`
 - 详情：是否启用图片尺寸支持。
 
-### obsidianImgSize
+### legacyImgSize
```
Mister-Hope, line 192: same

plugins/markdown/plugin-markdown-image/src/node/markdownImagePlugin.ts
```diff
@@ -19,8 +19,8 @@ export const markdownImagePlugin = (
 
     if (options.figure) md.use(figure)
     if (options.lazyload) md.use(imgLazyload)
-    if (options.obsidianSize) md.use(obsidianImageSize)
     if (options.size) md.use(imgSize)
+    if (options.legacyImgSize) md.use(legacyImgSize)
```
Mister-Hope, line 23: same

plugins/markdown/plugin-markdown-image/src/node/options.ts
```diff
@@ -39,11 +39,11 @@ export interface MarkdownImagePluginOptions {
...
-   * Whether to enable obsidian image size mark support
+   * Whether to enable legacy image size mark support
    *
-   * 是否启用 obsidian 图片大小标记支持。
+   * 是否启用旧版图片大小标记支持。
    *
    * @default false
    */
-  obsidianSize?: boolean
+  legacyImgSize?: boolean
```
Mister-Hope, line 48: same

docs/.vuepress/configs/plugins.ts
LincZero: https://mdit-plugins.github.io/zh/img-size.html 
This link states "legacySize",
Are you sure different names should be used here.

Mister-Hope: Why not? Other options are not `imgSize` `imgMark`, we are creating a `markdown-image` plugin

Mister-Hope: This is not changed.

Mister-Hope:
CI is broken as well, I think you might forget a successful `pnpm i`

LincZero:
Both my local `pnpm i` and `pnpm build` were successful before committing. I think the file was missed just before committing

LincZero:
I don't understand. Directly trace back the lock file, and then re `pnpm i` on github codespace.

Mister-Hope (changes requested):
linter test failed

Mister-Hope (approved):

PR #453, opened 2025-05-06
feat(helper): provide slot types

Mister-Hope:


pengzhanbo:
I see that the `Slot` type is being used, but I couldn't find where its type declaration is located.

Mister-Hope:
You are right, I forgot to upload it.

PR #455, opened 2025-05-08
feat: add plugin-llms

pengzhanbo:

- [ ] Bug fix
- [x] New feature
- [ ] Other

### Screenshots

**Before**

**After**


.vscode/settings.json
```diff
@@ -49,6 +49,8 @@
     "lazyload",
     "lightmode",
     "linkify",
+    "llms",
+    "Llmstxt",
```
Mister-Hope, line 53:
```suggestion
    "llmstxt",
```

plugins/ai/plugin-llms/src/node/constants.ts
```diff
@@ -0,0 +1,13 @@
+export const PLUGIN_NAME = '@vuepress/plugin-llms'
+
+/** Default template for the `llms.txt` file. */
+export const DEFAULT_LLMSTXT_TEMPLATE = `\
+# {title}
```
Mister-Hope, line 5: Can we make the template variable a little bit complicated? E.g.: `${{title}}`

pengzhanbo: The template does not need to consider complex scenarios; `{foo}` can already accomplish the task of the template very well.

PR #463, opened 2025-05-10
fix(plugin-shiki): fix incorrect highlight lines parsing

pengzhanbo:

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Screenshots

**Before**

**After**


Mister-Hope (approved):
I am just approving it as I am not familiar with Shiki transformers.

It seems like highlight lines is already fully handled by an internal Shiki transformer?

If this is correct, you can just merge it after my approval.

pengzhanbo:
@Mister-Hope 
In fact, `highlight lines` does not directly conflict with `collapsed-lines`; it conflicts with any text that contains a `-`. The regular expression matching and character replacement in the `highlightLinesPlugin` caused this issue, but it is actually meaningless and can be handled simply in `attrsToLines`.

PR #467, opened 2025-05-12
fix(theme-default): fix NotFound mismatch

Mister-Hope:


Mister-Hope:
Some reviews are needed here @meteorlxy @pengzhanbo.

A mismatch happens when entering a non-existing locale path. It will fall back to `/404.html`, which is rendered with root locale lang, but then reset the route to `/<locale>/404.html` while hydrating, this causes a SSR mismatch.

This issue has multiple valid fix, e.g.: wrapping the whole 404 pages with `<ClientOnly>`

I try to solve this issue by rendering all locale 404 pages with root locale page, then change the display language after onMounted. This will introduce a flash changing the display text if users are opening a path like `/zh/not-existed.html` directly, but with router actions everything will be fine.

I am not sure that this approve is the best one, but I believe a flash from root locale to current is better than an empty 404 page in SSR.

Any better ideas are appreciated.

PR #476, opened 2025-05-21
fix(plugin-copy-code): fix hydration mismatch

pengzhanbo:

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Screenshots

**Before**

**After**

### DIFF
--- a/plugins/features/plugin-copy-code/src/client/composables/useCopyCode.ts
+++ b/plugins/features/plugin-copy-code/src/client/composables/useCopyCode.ts
@@ -82,20 +82,19 @@ export const useCopyCode = ({
     codeBlockElement.setAttribute('copy-code', '')
   }
 
-  const appendCopyButton = async (): Promise<void> => {
-    await nextTick()
+  const appendCopyButton = (): void => {
     document.body.classList.toggle('no-copy-code', !enabled.value)
     if (!enabled.value) return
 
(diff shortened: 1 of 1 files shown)


plugins/features/plugin-copy-code/src/client/composables/useCopyCode.ts
```diff
@@ -93,8 +94,8 @@ export const useCopyCode = ({
     flush: 'post',
   })
 
-  onContentUpdated((reason) => {
-    if (reason !== 'beforeUnmount') appendCopyButton()
+  onContentUpdated(async (reason) => {
```
Mister-Hope, line 97: I am not sure, but are there actual benefits to add a async function here rather than `void appe....`?

Also, if nextTick is needed there, I suspect this should be an improvement for the onContentUpdate lifecycle.

When we use this, we absolutely need to access the updated dom immediately when reason is not `beforeunmount`

pengzhanbo: The actual issue does not come from `onContentUpdated`, but rather from `watchImmediate`, which leads to a hydration mismatch. 

it can be changed to:
```ts
watchImmediate(enabled, () => nextTick(appendCopyButton), {
  flush: 'post',
})
```

Mister-Hope: That make sense.

PR #478, opened 2025-05-23
fix(plugin-docsearch): open on ctrl+k when page first loads

xav-ie:
The `{ once: true }` prevented the listener from properly attaching
and listening for a whole ctrl+k event. It would exit early, causing
the page to not function properly on first load.

`useEventListener` automatically removes itself on unmount, so I am not sure the purpose of this line. I am hopeful this is not intentional and simple mistake. By approving and merging this PR, you will greatly improve my quality of docs searching. Otherwise, I will need custom scripts :/.

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Description

The search feature now opens with ctrl+k consistently.

### Screenshots

**Before**

https://github.com/user-attachments/assets/8ddc72b2-0172-4b51-9fc2-90dfc6193433

**After**

https://github.com/user-attachments/assets/0797b66e-459e-49e1-9be6-f78e90bf7dfa

--- 

The `before` branch adds the plugin inside the e2e thing. I am not sure if that is wanted. It requires some extra setup to test right now. I left that part out of this PR. 

I am hopeful that once this is merged in that ctrl+k will work on sites like https://www.nushell.sh! :D

### DIFF
--- a/plugins/search/plugin-docsearch/src/client/composables/useDocSearchHotkeyListener.ts
+++ b/plugins/search/plugin-docsearch/src/client/composables/useDocSearchHotkeyListener.ts
@@ -4,19 +4,15 @@ import { useEventListener } from '@vueuse/core'
  * Add hotkey listener, remove it after triggered
  */
 export const useDocSearchHotkeyListener = (callback: () => void): void => {
-  useEventListener(
-    'keydown',
-    (event) => {
-      const isHotKeyBind = event.key === 'k' && (event.ctrlKey || event.metaKey)
-      const isSlashKey = event.key === '/'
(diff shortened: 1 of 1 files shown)


Mister-Hope:
See https://github.com/algolia/docsearch/issues/1363.

I will leave this for now, as the `once: true` does fix some issues.

xav-ie:
Thank you, Mister-Hope, for pointing me in correct direction. It now makes sense why code has to be this way for now.

I was able to find your old PR there and make it return an unmounting function. You are free to close this or keep it open and I will just make it use the unmounting function once available.

PR #481, opened 2025-05-28
feat(helper): add useMounted

Mister-Hope:
This would be useful when SSR/CSR DOM can not guarantee to be the same, for example, generating a random ID for `aria-controls` attribute.


pengzhanbo:
There is already a similar implementation of [useMounted](https://vueuse.org/core/useMounted/) in `@vueuse/core`, and I think it can be used directly.

PR #499, opened 2025-06-03
refactor(plugin-toc): improve jsdoc and fix docs

Mister-Hope:
BREAKING CHANGES: rename `headerOptions` as `headersOptions`, `propsOptions` as `renderOptions`


Mister-Hope:
I believe it's necessary to make the 2 changes.

- `headersOptions` is now aligned in component props and plugin options
- `propsOptions` used to be the actual component props, but it's no longer the case after we switched to DOM-based header, so it's changed to `renderOptions`

PR #535, opened 2025-06-10
feat: use chokidar v4

Mister-Hope:


Mister-Hope:
@pengzhanbo Need some help here, register-components plugin is the only thing that blocks us from upgrading chokidar now.

Chokidar v3 is no longer maintained.

pengzhanbo (approved):

pengzhanbo (approved):

PR #563, opened 2025-10-14
chore: bump `vuepress` to `2.0.0-rc.26`

pengzhanbo:

- [ ] Bug fix
- [ ] New feature
- [x] Other

### Screenshots

**Before**

**After**


Mister-Hope:
I believe things has already been done in main, only docs needs to change.

Also Vue-router treats the missing file as bug and add it back as legacy.

PR #574, opened 2025-11-24
feat(plugin-markdown-chart)!: avoid potential XSS attack

Mister-Hope:
BREAKING CHANGES: echart variables in scripts are changed, script blocks MUST be manually allowed


Mister-Hope:
@pengzhanbo RFC here, I believe this is a necessary change, as most maintainers may not realize that certain JS code blocks from chart containers will be executed in browser, especially these functions are default enabled via certain templates.

We add 2 UPPER_CASE options for this and ensure only scripts from a whitelist can be run in site.

PR #621, opened 2026-03-02
fix(plugin-git): support submodule

mtfhx:

- [X] Bug fix
- [ ] New feature
- [ ] Other

### Description

When VuePress is built from a subdirectory or within a monorepo, plugin-git may fail to locate files correctly.

In my case, the documentation content of the site is maintained in an independent private repository and linked via Git submodules (the site itself is public, but the content repository is private). In this setup, the Git plugin is unable to correctly retrieve commit metadata such as author and date.

This PR improves the behavior with the following workflow:
	1. Detect the Git repository root using: `git rev-parse --show-toplevel`
	2. Resolve the target file path:
		• Convert to absolute path first
		• Then transform it into a path relative to the detected repository root
	3. Execute git log from the repository root directory instead of the current working directory

### Screenshots

**Before**

**After**


Mister-Hope (changes requested):
The Git plugin is slow enough. I believe a directory cache can be stored here. A directory shall be resolved only once.

e.g.: `/path/a.md` triggers a check for the `/path/` directory, and `/path/b.md` shall reuse the result from the cache later

Since extendsPage hooks are async, you may need to handle promises here in case the commands are still executed several times due to the delayed return of the first one.

Adding extra memory here shall be worthy for a cache, and the cache map shall be cleared in the onPrepared stage (to save build memory usage), as at this moment we can no longer add pages.

Mister-Hope:
It seems that copilot shares the same view with me.

A cache is necessary

Mister-Hope (commented):
I am not sure how the logger.error will be triggered, can you plz provide some examples? Could it be too noisy in some cases?

Mister-Hope (approved):

mtfhx:
I’ve tried to fix the link issue. @Mister-Hope @pengzhanbo 

Please review it when convenient. Thanks!

Mister-Hope:
Appreciated, I am not an expert at Git, so let me review it carefully across different agents with their explanations about these commands.

Mister-Hope:
@pengzhanbo Local security check suggests git plugin is now dangerous as filePath are passed to command args without validating.

It would be dangerous if the filePath is rewritten to something like `a.md && cat /etc/passwd` or `a.md && rm -rf /.` in functions like `getGitCreateDate`.

Could you check this part once this PR is merged? I will currently leave these security issue here.

Mister-Hope:
@mtfhx I fix a few bugs and refine some codes. You can review these changes I made.

I can not enter a valid github copilot pro subscription now. Now let's wait for other team members to trigger copilot review.

mtfhx:
Fine. I've done some simple testing locally for parallel submodules and nested sub-module scenarios, and no issues were found.

Waiting for Copilot review ...

Mister-Hope:
@meteorlxy Could you trigger a copilot review?

Mister-Hope (dismissed):

PR #672, opened 2026-05-06
build: bump deps

Mister-Hope:


Mister-Hope:
@copilot Run pnpm lint and fix the warnings gracefully. If this is a false positive, use `// oxlint-disable-next-line ruleName`

PR #680, opened 2026-06-04
fix(plugin-slimsearch): avoid crash when custom field formatter is no…

Checkmate544:
…t a string

The search result renderer called .split('$content') on the resolved custom-field formatter. When the formatter is a per-locale record whose value for the current locale is missing or not a string, .split throws, corrupting the render tree and blanking the whole result panel.

Guard the resolved value with isString so a non-string formatter falls back to an empty template instead of throwing.

- [x] Bug fix
- [ ] New feature
- [ ] Other

### Screenshots

**Before**

**After**

### DIFF
--- a/plugins/search/plugin-slimsearch/src/client/components/SearchResult.ts
+++ b/plugins/search/plugin-slimsearch/src/client/components/SearchResult.ts
@@ -145,11 +145,12 @@ export default defineComponent({
     const getDisplay = (matchedItem: MatchedItem): (VNode | string)[] => {
       if (matchedItem.type === 'customField') {
         const formatterConfig = customFieldConfig[matchedItem.index]
+        const formatter = isPlainObject(formatterConfig)
+          ? formatterConfig[routeLocale.value]
+          : formatterConfig
 
         const [prefix, suffix = ''] = (
-          (isPlainObject(formatterConfig)
-            ? formatterConfig[routeLocale.value]
-            : formatterConfig) ?? ''
+          isString(formatter) ? formatter : ''
         ).split('$content')
 
         return matchedItem.display.map((display) =>


Mister-Hope:
I refuse "Enhancement" like this.

The first thing is that you should provide "correct" options.

Unlike Webpack that introduce a config schema to validate whole options, vite ecosystem does not.

Try setting `root: 123` inside config file in any vite project, you will get CRASH — `ERR_INVALID_ARG_TYPE` error from Node's `path.resolve`. Vite won't convert to `'123'` for you. Same thing happens in Vue ecosystem (Vue createApp options, @vue/compiler-sfc, etc.)

plugins/search/plugin-slimsearch/src/client/components/SearchResult.ts
```diff
@@ -145,11 +145,12 @@ export default defineComponent({
     const getDisplay = (matchedItem: MatchedItem): (VNode | string)[] => {
       if (matchedItem.type === 'customField') {
         const formatterConfig = customFieldConfig[matchedItem.index]
+        const formatter = isPlainObject(formatterConfig)
+          ? formatterConfig[routeLocale.value]
+          : formatterConfig
 
         const [prefix, suffix = ''] = (
-          (isPlainObject(formatterConfig)
-            ? formatterConfig[routeLocale.value]
```
Mister-Hope, line 151: ~~It would be fine if this is changed to `formatterConfig[routeLocale.value] ?? formatterConfig['/']`, AKA falling back to root locale.~~ No that's not reasonable.

But no further changes are accepted.

We never validate any plugin options and assume them to be correct.

Mister-Hope:
Wait: https://github.com/vuepress/ecosystem/commit/0bf0a99f689af2da7f7319e53f91390bbf671cd4 should fix your original issue 2 weeks ago (before this PR is raised), yet this is not published.

PR #685, opened 2026-06-24
perf(plugin-llms): use async remark processing with parallel

Mister-Hope:


Mister-Hope:
@pengzhanbo please manually review this.

pengzhanbo (approved):

PR #787, opened 2026-08-14
feat: bump to vp2rc31

Mister-Hope:


Mister-Hope:
@pengzhanbo Your replace-assets plugin breaks, plz fix it.

This is now the only blocking thing that stops us moving forward.

Issue #603, opened 2026-02-09
[Feature request] plugin-code-tree


Mister-Hope:
### Clear and concise description of the problem

Gracefully display small templates.

Code tabs can do similar thing, but the file structure is not clear.

### Suggested solution

Similar to https://theme-plume.vuejs.press/guide/markdown/code-tree/

PR #849, opened 2026-09-10
feat(plugin-markdown-code-tree): add code tree plugin (#603)

Mister-Hope:


Mister-Hope:
I'm not sure if we should make this a separate plugin or integrate it into the existing markdown file tree plugin. I'm open to both approaches, especially considering the significant feature overlap between the file tree and the code tree.

